PR #1, opened 2018-09-13
Implement distributed bootstrap (bug 1485811)

robhudson:


Dockerfile
```diff
@@ -0,0 +1,19 @@
+# This Dockerfile supports the bin/test script for local testing.
+#
+# Make sure any changes here stay in sync with .circleci/config.yml
+# so local testing and CI are comparable.
+
+ARG PYTHON_VERSION=2.7
+FROM python:$PYTHON_VERSION
+
+RUN apt-get update && apt-get install -y libsnappy-dev openjdk-8-jre-headless
```
sunahsuh: As we discussed, we should check to see if we need snappy

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
+from functools import partial
+
+
+def bootstrap(iteration, stat_fn, broadcast_data):
```
sunahsuh: I'd rename this `_pseudosample` or something, and rename `run_bootstrap` to `bootstrap`

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
...
+    np.random.seed()
+    n = len(broadcast_data.value)
+    randints = np.random.randint(0, n, n)
+    return stat_fn(broadcast_data.value[randints])
+
+
+def run_bootstrap(sc, data, stat_fn, num_iterations=1000, confidence_level=0.95):
```
sunahsuh: Let's default to 2000 -- 1000 is actually a bit low

sunahsuh: Can we add docstrings to this? Doesn't have to be in this review, but we should generate docs before we call this done

tests/conftest.py
```diff
@@ -0,0 +1,29 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+import logging
+
+import pytest
+from pyspark.sql import SparkSession
+
+
+@pytest.fixture(scope="session")
```
sunahsuh: Consider using https://github.com/holdenk/spark-testing-base

tests/test_bootstrap.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_mean():
+    assert np.mean(DATA) == 4.125
+
+
+def test_bootstrap_mean(spark_context):
+    d = run_bootstrap(spark_context, DATA, np.mean, num_iterations=10)
+    assert d['calculated_value'] == 4.125
+    assert d['confidence_low'] < d['calculated_value'] < d['confidence_high']
```
sunahsuh: Good start -- is the plan to add a test using a normal distribution as we discussed?

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # Broadcast the data as read-only to the clusters.
+    broadcast_data = sc.broadcast(data)
+
+    f = partial(bootstrap, stat_fn=stat_fn, broadcast_data=broadcast_data)
+    stats = sc.parallelize(range(num_iterations)).map(f).collect()
+
+    broadcast_data.unpersist()
+
+    # Given the distribution of stats, calculate the percentiles based on the
```
sunahsuh: Actually, one more thought: consider abstracting getting the confidence intervals given a list of pseudostats into another method called `_ci_percentile` -- this way, if we decide to implement bca or some other algorithm we can make it easy to choose one of those as a parameter

mozanalysis/bootstrap.py
sunahsuh: Actually, maybe `_resample`

robhudson: Is `bootstrap` a good name? E.g. if we later add a `poisson_bootstrap`, would just plain-old `bootstrap` still make sense?

.circleci/config.yml
```diff
@@ -0,0 +1,108 @@
...
+
+  py27:
+    <<: *test_settings
+    docker:
+      - image: python:2.7
+
+#  py37:
+#    <<: *test_settings
+#    docker:
+#      - image: python:3.7
```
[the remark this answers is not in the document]
robhudson: I don't see anything right now that would cause me to think otherwise. I'll test it and if so, uncomment this.

mozanalysis/bootstrap.py
sunahsuh: I think if we add poisson, we'd make it a parameter (maybe called `resampling_method`) to a `bootstrap` method

tests/test_bootstrap.py
robhudson: Yes, I forgot about that. Thanks for the reminder.

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
+from functools import partial
+
+
+def bootstrap(iteration, stat_fn, broadcast_data):
+    np.random.seed()
```
sunahsuh: One more thing that I forgot about -- if you seed this with the iteration number (plus maybe some other characteristic of the original sample like, n), you can make this deterministic which would be nice for repeatability

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def _resample(iteration, stat_fn, broadcast_data):
+    np.random.seed()
+    n = len(broadcast_data.value)
+    randints = np.random.randint(0, n, n)
+    return stat_fn(broadcast_data.value[randints])
+
+
+def bootstrap(sc, data, stat_fn, num_iterations=2000, confidence_level=0.95):
```
jklukas: Is documentation of the API going to be a separate step? I'd love to see a docstring here explaining the purpose of the function and describing the parameters (in numpy docs style if you don't have another standard for this project).

mozanalysis/bootstrap.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
```
jezdez: I recommend to move the source into a `src/` directory to prevent import time issues when the package is located in the root directory and overlaps with installed packages (e.g. during coverage collection). A good primer is located here: https://hynek.me/articles/testing-packaging/

mozanalysis/bootstrap.py
robhudson: Yes I'd like to add docstrings. I'll take a look at numpy. Thanks for the suggestion.

jklukas: See https://numpydoc.readthedocs.io/en/latest/

To my knowledge, we tend to use numpy-style docstrings. They're pretty widely used, especially in python data world.

jezdez: FWIW, I've stayed close to standard Sphinx/rST docstring syntax for parameters etc in other data specific projects but I could have missed precedence in some cases.

jklukas: @jezdez has much better context than I on precedent for Python at Mozilla, so do follow his advice here. numpydoc may well be overkill here.

src/mozanalysis/bootstrap.py
```diff
@@ -21,6 +21,24 @@ def _confidence(data, confidence_level):
     }
 
 def bootstrap(sc, data, stat_fn, num_iterations=2000, confidence_level=0.95):
+    """Returns a bootstrapped confidence interval
+
+    Given an array of data, returns the bootstrapped confidence interval based
+    on the confidence level provided. This uses pyspark and distributes the data
+    to the nodes for efficiency.
```
sunahsuh: I'd note the limitations: all the data plus one replicate should be able to fit in memory

src/mozanalysis/bootstrap.py
```diff
@@ -9,6 +9,17 @@ def _resample(iteration, stat_fn, broadcast_data):
     return stat_fn(broadcast_data.value[randints])
 
 
+def _confidence(data, confidence_level):
```
sunahsuh: I'd rename this `_percentile`

mozanalysis/bootstrap.py
robhudson: Thanks all for the feedback.

tests/conftest.py
robhudson: I did take a look but felt that the Python implementation doesn't do much more than what's being done with the pytest fixtures. If later we think it's beneficial we can always switch over.

PR #9, opened 2018-10-25
Add ExperimentAnalysis and engagement metrics

robhudson:


src/mozanalysis/metrics.py
```diff
@@ -0,0 +1,121 @@
...
+    "SUM(scalar_parent_browser_engagement_total_uri_count)"
+).alias("sum_total_uris")
+
+sum_total_hours_expr = F.expr(
+    "SUM(subsession_length / {seconds_per_hour})".format(**CONSTANTS)
+).alias("sum_total_hours")
+
+sum_active_hours_expr = F.expr(
+    "SUM(active_ticks * {ticks_per_second} / {seconds_per_hour})".format(**CONSTANTS)
+).alias("sum_active_hours")
```
sunahsuh: Thoughts on making some of the re-used column expressions in the definitions constants here? At the very least it means pyflake can catch typos

tests/test_experiments.py
```diff
@@ -0,0 +1,101 @@
...
+    ]
+    # Only calculate the means to reduce bootstrap time during testing.
+    for m in metrics:
+        m.stats = [np.mean]
+
+    df = _generate_data(spark, {"aaaa": "control", "bbbb": "variant"})
+    pdf = ExperimentAnalysis(df).metrics(*metrics).run()
+
+    # Just get the control rows.
+    pdf = pdf[pdf.branch == "control"]
```
sunahsuh: We should probably check there's actually data for the other branch too :)

src/mozanalysis/experiments.py
```diff
@@ -0,0 +1,110 @@
...
+                        stat,
+                    )
+                    data.append(
+                        {
+                            "branch": split,
+                            "metric_name": metric_name,
+                            "stat_name": stat.func_name,
+                            "stat_value": bs["calculated_value"],
+                            "ci_low": bs["confidence_low"],
+                            "ci_high": bs["confidence_high"],
```
sunahsuh: Now that I'm looking at this, some of the other info we'd probably want for each of these is:
how many distinct clients and how many client-days contributed to calculating this metric?
Not a comment for this round, but something to note for future work

tests/test_experiments.py
```diff
@@ -0,0 +1,101 @@
...
+    # Note: 35.625 is the sum of the active ticks * 5 / 3600 over 9 days.
+    assert stat_values[1] == 35.625 / 9
+
+    # engagement_uris_per_active_hour
+    # Note: 342 is the sum of the URIs, 35.625 is the sum of the active hours.
+    assert stat_values[2] == 342 / (1 / 3600.0 + 35.625)
+
+    # engagement_intensity
+    # Note: 4.75 is the sum of the total hours, 35.625 is the sum of the active hours.
+    assert stat_values[3] == 4.75 / (1 / 3600.0 + 35.625)
```
sunahsuh: I'd definitely like more unit tests of the `aggregate_per_client_daily` and the `analyze` methods, but I don't think that needs to be blocking for this PR

src/mozanalysis/metrics.py
```diff
@@ -0,0 +1,121 @@
...
+        `pyspark.sql.functions`, e.g. `F.sum`, to perform the per-client
+        aggregate on the columns defined in `columns`.
+    final_expression (`pyspark.sql.function.expr`) : An expression used as a
+        final computation against the aggregated columns.
+
+    Subclasses can optionally define the following:
+
+    stats (list of executables) : A list of executables passed to `bootstrap`
+        for computing the statistics. The default is stats for the 5th, 25th,
+        50th, 75th, and 95th percentiles.
```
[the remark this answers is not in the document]
robhudson: Oops, you're right. I need to add the 50th.

src/mozanalysis/metrics.py
robhudson: Which ones are you referring to?

sunahsuh: e.g. `F.col("subsession_length")`

src/mozanalysis/utils.py
```diff
@@ -0,0 +1,22 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import pyspark.sql.functions as F
+from pyspark.sql.column import Column
+
+
+def dedupe_columns(columns):
+    """
+    Given a list of columns, returns a list with duplicates removed.
+
+    These can be either from `F.col` or `F.expr`.
```
sunahsuh: I think a longer explanation of why we need to do this instead of using `set` would be useful, including the context with test failing on python 3, in the off chance the underlying assumptions change in the future

PR #16, opened 2018-11-21
Handle null values

tdsmith:
When I tried running this on data from an in-the-wild experiment I got some failures about missing values.
This adds a test and then handles the case where clients don't report values.
For all of these values, the true value should be zero when the observation is missing.

### DIFF
--- a/src/mozanalysis/metrics.py
+++ b/src/mozanalysis/metrics.py
@@ -61,15 +61,19 @@ class MetricDefinition(object):
 
 
 sum_total_uris_expr = F.expr(
-    "SUM(scalar_parent_browser_engagement_total_uri_count)"
+    "SUM(COALESCE(scalar_parent_browser_engagement_total_uri_count, 0))"
 ).alias("sum_total_uris")
 
+
 sum_total_hours_expr = F.expr(
-    "SUM(subsession_length / {seconds_per_hour})".format(**CONSTANTS)
+    "SUM(COALESCE(subsession_length, 0) / {seconds_per_hour})".format(**CONSTANTS)
 ).alias("sum_total_hours")
 
+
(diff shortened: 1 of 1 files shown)


tests/test_experiments.py
```diff
@@ -172,3 +173,33 @@ def test_engagement_metrics(spark):
...
+    data = {
+        "client_id": ["a", "b", "c"],
+        "experiment_branch": ["control"] * 3,
+        "submission_date_s3": ["1", "2", "3"],
+        "scalar_parent_browser_engagement_total_uri_count": [None, 10, 10],
+        "subsession_length": [10, None, 10],
+        "active_ticks": [10, 10, None],
+    }
+    sdf = spark.createDataFrame(pd.DataFrame(data, dtype=object))
+    ExperimentAnalysis(sdf).metrics(*metrics).run()
```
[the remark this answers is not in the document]
tdsmith: Okay, added some additional assertions about the output.

src/mozanalysis/metrics.py
```diff
@@ -60,16 +60,20 @@ class MetricDefinition(object):
     stats = [np.mean, p5, p25, p50, p75, p95]
 
 
-sum_total_uris_expr = F.expr(
-    "SUM(scalar_parent_browser_engagement_total_uri_count)"
+sum_total_uris_expr = F.coalesce(
+    F.sum("scalar_parent_browser_engagement_total_uri_count"), F.lit(0)
```
robhudson: I'm curious why you didn't use the similar `COALESCE` pattern here as well?

robhudson: (Inside the `F.expr` I mean.)

PR #17, opened 2018-11-21
Make the tests a little more concise?

tdsmith:
I did this to help me understand the tests -- originally I thought I was going to address #16 in part by adding rows to _generate_data, but then decided against. I don't have strong feelings about whether this is a better way to write the tests or not. I thought it might be a little clearer but maybe that's just because rewriting them forced me to understand them. If you don't prefer this then let's not bother changing it.


robhudson:
This looks great to me! Could we move some of the comments into the green if you agree they are helpful?

robhudson:
This now has conflicts. Once fixed feel free to merge.

tdsmith:
I don't have write permissions but this isn't conflicty anymore.

PR #18, opened 2018-11-22
Add some UDFs

### LINKED ISSUE #10: Add helpful UDFs
There's a few pyspark UDFs I keep writing to handle histogram columns in main_summary or rehydrated from pings, like `histogram_sum` or `histogram_count`. I'd also like to compute histogram quantiles. I think this is could be a helpful place for them to live.  @robhudson, do you agree?

tdsmith:
Closes #10.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,14 @@ setup(
     url='https://github.com/mozilla/mozanalysis',
     packages=find_packages(where='src'),
     package_dir={'': 'src'},
-    install_requires=['py4j', 'numpy', 'pandas'],
+    install_requires=[
+        'py4j',
+        'numpy',
+        'pandas',
+        'pyarrow>=0.8.0',
+        'scipy',
+        'statsmodels>=0.9',
+    ],
     setup_requires=['pytest-runner', 'setuptools_scm'],
     extras_require={
         'testing': tests_require,
--- a/src/mozanalysis/udf.py
+++ b/src/mozanalysis/udf.py
@@ -1,0 +1,121 @@
+from pyspark.sql import functions as F
+from pyspark.sql.functions import udf
+from pyspark.sql.types import DoubleType, LongType, StructField
+
+from statsmodels.stats.weightstats import DescrStatsW
+
+
+@udf(LongType())
+def histogram_sum(values):
+    """Returns the weighted sum of the histogram.
+
+    Note that this is the sum *post*-quantization, which amounts to a
+    left-hand-rule discrete integral of the histogram. It is therefore
+    likely to be an underestimate of the true sum.
+    """
+    if values is None:
+        return None
+    return sum(int(k)*v for k, v in values.items())
+
+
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -6,21 +6,27 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from setuptools import setup, find_packages
 
-tests_require = ['mock', 'pyspark', 'pytest', 'pytest-cov', 'pytest-timeout']
+tests_require = ["mock", "pyspark", "pytest", "pytest-cov", "pytest-timeout"]
```
tdsmith: Oh, sorry for the quote changes; I ran `black` over it and didn't look at the output.

src/mozanalysis/udf.py
```diff
@@ -0,0 +1,121 @@
...
+from statsmodels.stats.weightstats import DescrStatsW
+
+
+@udf(LongType())
+def histogram_sum(values):
+    """Returns the weighted sum of the histogram.
+
+    Note that this is the sum *post*-quantization, which amounts to a
+    left-hand-rule discrete integral of the histogram. It is therefore
+    likely to be an underestimate of the true sum.
```
tdsmith: It would be interesting to explore what we'd need to have in order to either a) make this a trapezoidal-rule integration or b) aggregate the high-resolution histogram sums that the clients report.

For (a), we'd have to know the name of the column and generate the list of bins. This is maybe out of scope right now.

tests/test_udf.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        .groupBy(sdf.client_id)
+        .apply(threshold_udf)
+        .toPandas()
+        .set_index(["client_id", "threshold"])
+    )
+
+    def check(client, threshold, fraction):
+        value = df.loc[(client, threshold), "fraction_exceeding"]
+        assert abs(value - fraction) < 1e-6
```
robhudson: You could use `[pytest.approx](https://docs.pytest.org/en/latest/reference.html#pytest-approx)` here if you'd like, e.g.: `assert fraction == approx(value)`

setup.py
robhudson: No problem. I consider it good if this matches the black syntax but it is excluded in my `bin/format-python` script. If you'd prefer we can remove it from this PR and I can push the change myself.

robhudson:
This looks really great! For some reason I don't see that CircleCI picked this up. Perhaps try to force-push and see if it triggers a test run? I think the flake8 check will complain about the no spaces around operators with things like `int(k)*v`.

setup.py
tdsmith: I reverted and made the minimal change.

tdsmith:
That may be a config setting --

> By default, CircleCI does not build PRs from forked repositories. To change this setting, go to the Advanced Settings of your project and set the Build forked pull requests option to On.

PR #19, opened 2018-12-05
Fix intensity definition

tdsmith:
Per e.g.
https://metrics.mozilla.com/protected/sguha/shield_bootstrap.html,
intensity should be active_hours / total_hours, so that it ends up
generally being a fraction on [0, 1].


robhudson:
Thanks for the review on that! Could you also update the test that got this incorrect as well?

tdsmith:
I can indeed! Just wanted to send the fixup PR before I forgot.

robhudson:
This looks good. Thanks for catching it!!

PR #25, opened 2019-01-22
Add `mozanalysis.stats` to docs 

robhudson:


README.md
```diff
@@ -55,3 +55,24 @@ the release. Valid versions numbers are:
 Once the (signed) Git tag has been pushed to the main GitHub repository using
 git push origin --tags, Circle CI will automatically build and push a release to
 PyPI after the tests have passed.
+
+## Building the documentation
+
+The docs are located in the `docs/` directory. On each commit to the master
+branch the docs are built and pushed to Github pages.
+
+There are two ways to test the docs locally during composition:
```
tdsmith: Three, with `tox -e docs` :)

I guess I'm curious why we'd want to run these in docker instead of relying on tox for isolation, since we have to maintain a tox configuration anyway; it seems like there's less overhead if we just use tox.

setup.py
```diff
@@ -18,10 +18,11 @@
     packages=find_packages(where='src'),
     package_dir={'': 'src'},
     install_requires=[
-        'py4j',
         'numpy',
         'pandas',
+        'py4j',
         'pyarrow>=0.8.0',
+        'pyspark',
```
tdsmith: Is this safe / why did we avoid this before? (Were there concerns about pip's eager upgrade behavior?)

README.md
robhudson: That's assuming the user has `tox` locally. I do not. And when I tested this inside the docker container it requires `virtualenv` to be set up to run `tox`, which it is not. I think that's why I worked around it.

But good call out. I can investigate adding `tox` and `virtualenv` to the docker container so there's less ways to do this.

docs/conf.py
```diff
@@ -12,9 +12,18 @@
...
+sys.path.insert(
+    0,
+    os.path.normpath(
+        os.path.join(
+            os.path.abspath('.'),
+            '..',
+            'src',
+        )
+    )
+)
```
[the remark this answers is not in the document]
robhudson: Yes, which means I can remove the editable pip install. Thanks for catching that.

setup.py
robhudson: I don't recall a reason why this wasn't included. This project does require it to work.

docs/conf.py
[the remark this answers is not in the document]
robhudson: Turns out this isn't needed when using `tox` because it installs the project in the virtualenv. I removed it.

README.md
robhudson: I removed this documentation since `tox -e docs` is so easy. Unfortunately b/c of my personal computer set up I cannot use that, but no reason to add all this extra just for my unique case.

PR #26, opened 2019-01-29
Add the `contrib` namespace, and two util fns.

felixlawrence:
The contrib namespace is a place for code that is being re-used
across multiple reports/investigations, and may or may not be ready
for others to start using it yet. One day this code dreams of being
promoted out of contrib and into the main package. Until then,
others are welcome to look at and even use your code, but should
not be surprised when breaking changes are silently made. If they
want reliability, then they should ask for/help with its promotion
out of contrib.

### DIFF
--- a/src/mozanalysis/utils.py
+++ b/src/mozanalysis/utils.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import pyspark.sql.functions as F
 from pyspark.sql.column import Column
+from functools import reduce
 
 
 def dedupe_columns(columns):
@@ -26,3 +27,19 @@ def dedupe_columns(columns):
         d[col._jc] = col
 
     return list(d.values())
+
+
+def all_(l):
+    """Return the logical AND of the input list l.
+
+    Unlike the built-in `all`, this is compatible with dataframe columns.
+    """
+    return reduce(lambda x, y: x & y, l)
+
+
+def any_(l):
+    """Return the logical OR of the input list l.
+
+    Unlike the built-in `any`, this is compatible with dataframe columns.
+    """
+    return reduce(lambda x, y: x | y, l)
--- a/src/mozanalysis/contrib/__init__.py
+++ b/src/mozanalysis/contrib/__init__.py
@@ -1,0 +1,3 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/src/mozanalysis/contrib/flawrence/__init__.py
+++ b/src/mozanalysis/contrib/flawrence/__init__.py
@@ -1,0 +1,3 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/src/mozanalysis/contrib/flawrence/mobile.py
+++ b/src/mozanalysis/contrib/flawrence/mobile.py
@@ -1,0 +1,110 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from mozanalysis.utils import all_
+
+
+# An easy way to filter for particular event types
+metric_library = {
+    'session_start': {
+        'category': 'action',
+        'method': 'foreground',
+        'object': 'app',
+    },
+    'session_end': {
+        'category': 'action',
+        'method': 'background',
+        'object': 'app',
+    },
+    'erase_button': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'erase_button',
+    },
+    'back_button': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'back_button',
+    },
+    'notification': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'notification',
+    },
+    'notification_erase_open': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'notification_action',
+        'value': 'erase_open',
+    },
+    'home_shortcut': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'notification_action',
+        'value': 'erase_open',
+    },
+    'tabs_tray_erase': {
+        'category': 'action',
+        'method': 'click',
+        'object': 'tabs_tray',
+        'value': 'erase',
+    },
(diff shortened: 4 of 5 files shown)


felixlawrence:
Will probably want to skip coverage for `contrib` because it will hide the coverage of the part of the code that we really care about?

src/mozanalysis/utils.py
```diff
@@ -26,3 +27,19 @@ def dedupe_columns(columns):
         d[col._jc] = col
 
     return list(d.values())
+
+
+def all_(l):
+    """Return the logical AND of the input list l.
```
tdsmith: Well... bitwise, for primitives, or elementwise for certain collections -- maybe just "combines the elements of the iterable `l` using the `&` operator"?

Do you mind using Google docstrings? https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/mozanalysis/utils.py
```diff
@@ -26,3 +27,19 @@ def dedupe_columns(columns):
...
+
+    Unlike the built-in `all`, this is compatible with dataframe columns.
+    """
+    return reduce(lambda x, y: x & y, l)
+
+
+def any_(l):
+    """Return the logical OR of the input list l.
```
tdsmith: same concern here

tdsmith:
> Will probably want to skip coverage for contrib

Yes; please add an exclusion to the coverage config at https://github.com/mozilla/mozanalysis/blob/master/setup.cfg#L16

robhudson:
I'd like to discuss and decide on the `contrib/` flow and expectations. Apologies if this holds up this PR in the meantime but I think it's good to shake out now and then I can document it with the repo for future `contrib/` contributions...

> One day this code dreams of being promoted out of contrib and into the main package.

One worry I have is what if it never does? I imagine there would be some cases where the code doesn't evolve and just sits around. At what point do we decide to either promote the code or remove the code?

> Until then, others are welcome to look at and even use your code...

This is one reason why I think the personal namespacing is not advantageous. It sends the message of "don't look in here, this is mine only". Whereas I think the code *should* be looked at and attempted to be used to help it evolve outside of the contrib namespace.

tdsmith:
>> One day this code dreams of being promoted out of contrib and into the main package.
> One worry I have is what if it never does? I imagine there would be some cases where the code doesn't evolve and just sits around. At what point do we decide to either promote the code or remove the code?

I think the only stable answer to the latter question is "never," which I think is fine -- I'm happy with using mozanalysis as a vehicle for people to get their own utility functions onto clusters, and then users can make their own decisions about what code is worth retaining or promoting.

In the limit where nothing else ever gets promoted, I think I agree that's a partial failure, but I'm not too concerned about it since mozanalysis will still have/already has utility.

>> Until then, others are welcome to look at and even use your code...
> This is one reason why I think the personal namespacing is not advantageous. It sends the message of "don't look in here, this is mine only". Whereas I think the code should be looked at and attempted to be used to help it evolve outside of the contrib namespace.

I'd worry that using shared namespaces makes it less clear who has a) ownership and responsibility for maintaining or promoting code, and b) authority to make breaking changes. A tragedy of the commons!

I think a suggestion that code should be used by other humans also requires that things should have reasonable documentation, which for better or worse I think Felix is trying to avoid.

felixlawrence:
> I'd like to discuss and decide on the contrib/ flow and expectations. Apologies if this holds up this PR in the meantime but I think it's good to shake out now and then I can document it with the repo for future contrib/ contributions...

+1

I agree with Tim's points. Re: documentation - some level of documentation is definitely desirable; I certainly plan to clearly document any code that I think is ready and useful for others. I don't know whether we need enforcement of this though?

src/mozanalysis/utils.py
felixlawrence: Well, this is a question with surprising depth. I had a specific problem (the builtin `all` not playing nicely with spark dataframe columns) and wrote a one-liner to scratch that itch. But the one-liner is very general and could do many things other than scratch my original itch (the elements of the iterable don't need to be Columns). So I gave it a more general first line and demoted the reason for the function's existence (and probably the main reason it'll be used - otherwise people would just use the builtin) to the second line. Was this a mistake - should I have kept it front and center? "Return the logical AND of a list of `Column`s"?

"Combines the elements of the iterable `l` using the `&` operator" is a hair's breadth from "reduce over `l` using `&`"; i.e. it's just a restatement of the function's body and isn't particularly useful documentation IMO? Really people want to know whether this will solve their specific problem?

W.r.t. Google docstrings - are you asking for `Args` and `Returns` sections? I can do that - anything else my eye missed in the spot-the-difference?

robhudson:
> I'm happy with using mozanalysis as a vehicle for people to get their own utility functions onto clusters, and then users can make their own decisions about what code is worth retaining or promoting.

> I'd worry that using shared namespaces makes it less clear who has a) ownership and responsibility for maintaining or promoting code, and b) authority to make breaking changes. A tragedy of the commons!

I consider code added her as a shared resource. When it's added to mozanalysis, it's everyone's responsibility to make sure it's useful and is maintained. Person A could be actively doing analysis that requires a snippet but later move on to other tasks and person B needs that snippet and perhaps improves upon it or fixes an issue. I guess I see it as whoever decides to own it, owns it, whether that's multiple people working together on it or one person. What if person B wants to improve on person A's code snippet, would they copy it to their own personal namespace and make the change? Then there's 2 snippets that do almost identical things.

felixlawrence:
> What if person B wants to improve on person A's code snippet, would they copy it to their own personal namespace and make the change? Then there's 2 snippets that do almost identical things.

Yeah, this is annoying and triggering, so hopefully at least one of person A or B would feel uncomfortable with the situation, and they'd work together on a version of the code that fits both their needs and promote it out of `contrib`. Or person B would send person A a pull request. But if person A didn't want to co-operate with person B then at least person B would be able to plow ahead with unilateral changes in their own namespace without treading on person A's toes.

src/mozanalysis/utils.py
```diff
@@ -26,3 +27,37 @@ def dedupe_columns(columns):
...
+    Think of this as a vector-friendly version of the built-in function
+    `all()`.
+
+    Args:
+        l: A list of `Column`s of booleans. Or, more generally,
+            an iterable of items to reduce over.
+
+    Returns:
+        A `Column` of booleans representing the logical AND. Or,
+            more generally, the result of the logical AND.
```
tdsmith: I'm using up my pedantry budget for the month on the last day -- but in the general case it's a bitwise AND (or OR), not a logical AND.

I can fix this up later as part of turning on API docs for this page

felixlawrence: Wow, all this time I've been using bitwise AND/OR on booleans or 0/1 ints, thinking it was logical AND/OR for ndarrays - and have been doing so because of technical limitations in Python: `&` is overloadable but `and` is not, so best practice is to just use bitwise operations to do logical operations. Ew.

PR #28, opened 2019-02-06
Tests refactor

robhudson:


tests/conftest.py
```diff
@@ -7,19 +7,37 @@
...
+    _spark_context = _spark_session.sparkContext
 
     logger = logging.getLogger("py4j")
     logger.setLevel(logging.ERROR)
 
-    yield spark
-    spark.stop()
+
+def pytest_unconfigure(config):
+    global _spark_session
```
tdsmith: The `global` here isn't necessary since we aren't trying to assign to the name; same for the similar declarations in `spark()` and `spark_context()` -- I don't mind them if you think they're useful as signposting

tests/test_experiments.py
```diff
@@ -8,9 +8,14 @@
 import pandas as pd
 import pyspark.sql.functions as F
 
-# NOTE: The metrics are not imported here b/c they are evaluated and require a
-# spark context, which isn't available outside the test functions.
```
tdsmith: 🎉

tests/conftest.py
robhudson: Good catch. This file went through so many variations while I was testing what would work with pytest. I'll update and merge. Thanks!

PR #29, opened 2019-02-07
Proposed changes to support retention metrics

robhudson:


robhudson:
> I'm also uneasy about defining new object types that don't have any instance state; often it seems like the subclasses could be instances of the parent class instead. Maybe this is just an aesthetic preference.

Yes, the code here should have been better explained or documented. I think once you see the `RetentionMetrics` implemented you'll see what I'm going for. I should have that next week.

In the case of the `CoreMetric` which is expected to have a callables defined, being a subclass has a nicer code feel. Also, attaching methods to existing object instances is generally frowned upon. Compare the following:

```
# As a subclass
class MyMetric(CoreMetric):
    metrics = [...]

    def pre_callable(self, dataset):
        # Pre computation aggregation code here.
        pass
```

```
# As an instance
import types

def pre_callable(self, dataset):
    # Pre computation aggregation code here.
    pass

MyMetric = CoreMetric(metrics=[...])
MyMetric.pre_callable = types.MethodType(pre_callable, MyMetric)

```

There are more ways to do the above, all a little funky, but I think this demonstrates the reasoning of why I preferred subclasses for this case.

tdsmith:
I agree that's awkward; the pattern I've seen for handing a function to an object instance would be something like:

```python
class Metric:
  def __init__(self, mutators):
    self._mutators = mutators

  def do_a_thing(self, some_object):
    for f in self._mutators:
      some_object = f(some_object)
    return some_object

m = Metric(mutators=[lambda x: x+1])
```

robhudson:
Closing in preparation for a new PR with a different approach.

PR #31, opened 2019-03-12
Add search ads metric

benmiroglio:
This creates a contributor directory for my personal use (`contrib/bmiroglio`) containing a single function that calculates ad aggregates (sums and click through rates).


benmiroglio:
@felixlawrence Please take a look when you have a chance. I added a single function to make things simple, the main goal is getting my personal namespace established.

PR #37, opened 2019-04-03
Add retention metrics

### LINKED ISSUE #27: Implement retention metrics

robhudson:
This involves a major refactor to support both engagement and retention
metrics.

Fixes #27


robhudson:
@tdsmith Requested your review for initial feedback. There are a few TODOs in code that I'd like your opinion on also before fleshing those out.

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
+        self._df = value
+
+    def get_experiment_data(self):
+        df = self.spark_session.read.parquet("s3://telemetry-parquet/experiments/v1")
+        df = df.filter(F.col("experiment_id") == self.experiment_id)
+        return df
+
+    def aggregate_per_day(self, df, core_metric=None):
+        # If user wants to override outlier percentiles, set it False here and
+        # call it explicitly.
```
tdsmith: ?

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
+
+        return self.get_data()
+
+    @df.setter
+    def df(self, value):
+        self._df = value
+
+    def get_experiment_data(self):
+        df = self.spark_session.read.parquet("s3://telemetry-parquet/experiments/v1")
+        df = df.filter(F.col("experiment_id") == self.experiment_id)
```
tdsmith: It should be possible for the caller to specify this filtered table, perhaps as an argument to the constructor.

Using `df` for the name of this object is also a little confusing since class instances have a `df` property and this isn't that

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
+    def trim_outliers(self, outlier_percentile=0.9999, relative_error=0.0001):
+        # TODO
+        # Updates `self._df` and returns it.
+        return self._df
+
+    def engagement(self):
+        # Compute only the engagement metrics.
+        # Return pandas dataframe with data and CIs.
+        if not self._df:
+            self._df = self.get_data(metric=CoreMetrics.ENGAGEMENT)
```
tdsmith: What if `self._df` has the wrong metrics?

(Suggestion: either store the flag that `get_data` was passed and make a decision based on that, or else make calculating retention the responsibility of a second class)

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
-                        m.final_expression.alias(metric_name),
-                    ]
-                )
+        return agg_df
+
+    def retention(self, week=3):
+        # Compute only the retention metrics.
+        # Return pandas dataframe with data and CIs.
+        if not self._df:
+            self._df = self.get_data(metric=CoreMetrics.RETENTION)
```
tdsmith: What if `self._df` has the wrong metrics? (as above)

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
+                F.expr("CASE WHEN SUM(sum_total_hours) > 0 THEN 1 ELSE 0 END").alias(
+                    "retention"
+                ),
+                F.expr("MAX(CASE WHEN sum_total_uris >= 5 THEN 1 ELSE 0 END)").alias(
+                    "active_retention"
+                ),
             )
+        )
+
+        return week3_df
```
tdsmith: I don't think this tells me what I want to know, which is "what fraction of users that were present in week 0 are still present (and/or active) in week _n_"; this gives me a list of client_ids that were present in week 3?

tdsmith: I think you may be trying to produce something to bootstrap over? Filtering to the _n_th week means that this isn't a useful thing to bootstrap over (instead of joining a column onto the list of clients that represents whether they were present or active in week _n_, which would be) but also we can use the normal approximation for a binomial confidence interval, and save ourselves some computing: https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Normal_approximation_interval

tdsmith: (this difference may be another good reason to segregate the engagement and retention metrics)

src/mozanalysis/experiments.py
robhudson: What do you mean by "this filtered table"? The s3 parquet path?

src/mozanalysis/experiments.py
```diff
@@ -1,165 +1,268 @@
...
+
+    def analyze(self):
+        # TODO
+        # Compute all metrics.
+        #
+        # This is a convenience method.
+        #
+        # The idea would be to make the calls to `engagement()` and `retention`
+        # separately, pass those results to `bootstrap`, the concatenate the
+        # pandas dataframes returned from bootstrap and return that.
```
robhudson: @tdsmith This is one of my questions... whether this would be convenient and useful or not?

src/mozanalysis/experiments.py
robhudson: Thanks for pointing this out. At one point this code was the Python equivalent of sguha's SQL, but with multiple refactorings something must have gotten lost along the way. I'll go back and review this.

src/mozanalysis/experiments.py
tdsmith: I would like to be able to, at least optionally, pass in a Spark DataFrame which contains the rows of a table matching the `experiments` schema that are relevant to the experiment, instead of letting the class build that SDF itself from its own knowledge.

src/mozanalysis/experiments.py
tdsmith: Maybe a nice-to-have; I don't think it's essential.

src/mozanalysis/experiments.py
robhudson: Okay. That's what I had intended for the property setter on the `_df` variable. If you look at the tests it uses this to assign a dataframe for testing. The same could be done in a notebook, e.g., to assign the dataframe and still call `.engagement()` on it. Does that satisfy that use case?

tdsmith: That would work except that `ExperimentAnalysis.df` holds the result of `get_data`, which is a different data frame which has already been aggregated to the daily level, so it doesn't satisfy this use case.

robhudson: Can you elaborate on what the API would look like? Is this a requirement for this PR? I'm having trouble understanding exactly what you're expecting. The `get_data` call calls `get_experiment_data` and then `aggregate_per_day`. If you're saying that you want to provide a SDF that gets aggregated per day, that could be done via something like the following:

```
ea = ExperimentAnalysis(spark, "experiment-slug")
agg_df = ea.aggregate_per_day(my_sdf)
ea.df = agg_df
ea.retention()
...
```

src/mozanalysis/experiments.py
robhudson: I like the idea of storing some extra metadata about what's saved in the `_df`.

src/mozanalysis/experiments.py
robhudson: It appears that my Python is roughly equivalent to the SQL with one exception... at the end the SQL does a JOIN of the retention data with the engagement metrics. The subtlety that was missed was that any rows now in the retention table would JOIN with a zero in the retention columns. That is lacking from my implementation.

A proposed solution is to calculate each week's retention as separate columns, e.g. :`retained_in_week_1`, `retained_in_week_2`, `retained_in_week_3`, and likewise for active retention, so that the equivalent comparison could be made and also satisfy Tim's original comment in this thread.

Thoughts?

src/mozanalysis/experiments.py
robhudson: The code changed and the comment didn't. I'll remove the comment.

setup.py
```diff
@@ -18,6 +18,7 @@
     packages=find_packages(where='src'),
     package_dir={'': 'src'},
     install_requires=[
+        'enum34',
```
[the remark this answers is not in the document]
robhudson: Yes, good call. 👍

robhudson:
Closing this PR as I was told this is no longer needed.

PR #38, opened 2019-04-25
First step in a new direction.

felixlawrence:
This commit removes a lot of the existing code, and takes the first step on a new roadmap that takes mozanalysis beyond the `experiments` table, and aims to make safe and rigorous analyses easier than the alternative, while maintaining 'enough' flexibility.

This commit introduces the basic query infrastructure - most of the code that builds Spark queries.

Still to come in mozanalysis or mozreport (out of scope for this PR):
- A variety of stats functions to analyze the data output by `Experiment.get_per_client_data()`
- A library of pre-defined and blessed metrics
- Convenience functions for querying and analyzing time series (time since enrollment)
- Functions to visualize/present the output of stats functions

The plan is to leave this PR open until at least May 1, so that the DS team has time to review `experiment.py` against their needs, and request any major changes. Smaller suggestions are of course welcome at any time.


src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,509 @@
...
+                _Don't_ use `experiments`; as of 2019/04/02 it drops data
+                collected after people self-unenroll, so unenrolling users
+                will appear to churn. Must have at least the following
+                columns:
+                    - client_id (str)
+                    - submission_date_s3 (str)
+                    - data columns referred to in `metric_list`
+                Ideally also has:
+                    - experiments (map)
+            metric_list: A list of columns that aggregate and compute
```
harterrt: Can you note that these metrics will be applied over groups of (client, branch) data?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,509 @@
...
+                - ...
+                - [metric n]: The client's value for the nth (final)
+                    metric in `metric_list`.
+                - [sanity check 1]: The client's value for the first
+                    sanity check metric.
+                - ...
+                - [sanity check n]: The client's value for the last
+                    sanity check metric.
+
+            This format - the schema plus there being one row per
```
harterrt: Do we have some artifact of this decision?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,509 @@
...
+            enrolled client, regardless of whether the client has data
+            in `df` - was agreed upon by the DS team, and is the
+            standard format for queried experimental data.
+        """
+        req_dates_of_data = conv_window_start_days + conv_window_length_days
+
+        if self.num_dates_enrollment is None:
+            self._print_enrollment_window(last_date_full_data, req_dates_of_data)
+
+        # TODO accuracy: can/should we switch from submission_date_s3 to when the
```
harterrt: We made a decision re: activity vs submission date in [Bug 1422892](https://bugzilla.mozilla.org/show_bug.cgi?id=1422892). In short, we decided we should not trust activity date for business critical decisions.

If we want to reopen that discussion let's open a bug.

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,509 @@
...
+        elif end_date is not None:
+            enrollments = enrollments.filter(
+                enrollments.enrollment_date <= end_date
+            )
+
+        enrollments.cache()
+
+        return enrollments
+
+    def get_per_client_data(
```
harterrt: This seems reasonable, but for my own context, why not generate enrollments automatically?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,509 @@
...
+        we'll still want to include their data in the analysis, unless
+        we're explicitly using stats methods that handle censored data.
+
+
+    Example usage:
+        cd = spark.table('clients_daily')
+
+        experiment = Experiment(
+            experiment_slug='pref-flip-defaultoncookierestrictions-1506704',
+            start_date='20181215',
```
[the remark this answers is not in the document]
felixlawrence: IIUC the `experiments` table filters `clients_daily` based on the contents of the `experiments` map. At present, if a client does something that means they no longer match the normandy recipe, then they unenroll from the experiment and their future data won't appear in the `experiments` table - so they will appear to churn. This process is often branch-dependent and therefore is a source of bias. I can point you to a couple recent experiments this happened to (both of which were high volume so didn't appear in the `experiments` table anyway).

src/mozanalysis/experiment.py
felixlawrence: Heh, I was being a little cheeky here. Now (i.e. before this PR is merged) is an appropriate time to re-open the discussion if you would like to. [Here's the doc I took to the team meeting](https://docs.google.com/document/d/1S5_j6T-buAlQXqsevgGUvAD2Xt8qjeEjHfwc9xGhjiw/edit#heading=h.l1m5icicw80k) - we had a discussion and people seemed OK with it though there was no vote. If you or anyone else has concerns (including with the process) then please do speak up before next Wed; we should settle this issue before merging this code and text.

src/mozanalysis/experiment.py
felixlawrence: Awesome, thanks for the reference. I'll read up on it. Probably nothing's changed since that decision - maybe the sharper focus on conversion windows has some effect on the weights of things being traded off?

src/mozanalysis/experiment.py
felixlawrence: If you're looking at a conversion window with `conv_window_start_days > 0` then you can't guarantee that `df` has a row for each enrolled user (e.g. if the user didn't send a ping during the conversion window because they had already churned). So you already need to query enrollments then left join onto the data. Given that you have to do this, the `events` table is a better source of information because compared to the main ping there's less latency between enrolling and the event being sent, which means less chance for people to enroll without us learning about them.

src/mozanalysis/experiment.py
felixlawrence: Good point - done!

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
harterrt: I understand the unenrollment issue, but would that be an issue here? We're looking for the first date we saw data for any client, correct? If so, `experiments` should still work. IIUC, `experiments` has a row for each ping noting an active experiment.

src/mozanalysis/experiment.py
harterrt: Sounds good. We could also use this PR thread as the artifact of that decision if we are sure to get sguha and rweiss's explicit sign off as well.

src/mozanalysis/experiment.py
harterrt: Maybe, but I'm concerned this comment will get folks hyper-focused on the activity date vs submission date difference when they're trying to debug weird experiment results.

src/mozanalysis/experiment.py
harterrt: I think I'm confused. I understand why this function depends on a table of enrollments, but I'm interested in why we're passing this function a table of enrollments instead of calling `self.get_enrollments()` inside the function. This seems like a flexible way to build the API (which I like and agree with) but it does require some additional config and offers a chance to accidentally break `enrollments` before passing to `get_per_client_data`. 

Like I said, this seems reasonable as is. I'm mostly making sure I understand what this function is doing and that I'm not missing some nuance.

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
felixlawrence: Some experiments don't appear in `experiments` - e.g. high volume experiments, and I believe non-desktop experiments.

Also if we incorrectly filter dates so as to ignore the first 2 days of the experiment, then if we generate `enrollments` from enrollment events then we'll lose two days of enrolled users, but if we generate `enrollments` from `experiments` then a non-churned self-selected subset of these users will be included in `enrollments` with an incorrect enrollment date. I'd rather fail in the former fashion because the impact is more obvious and doesn't compromise the experiment by introducing potential bias.

src/mozanalysis/experiment.py
felixlawrence: Yes, this is here for flexibility. E.g. this week Megan needed to re-analyze an old mobile experiment, and was able to build her own `enrollments` dataframe and feed that in.

Eventually I think we should support segmenting too, and I think that would involve filtering the `enrollments` dataframe before feeding it into `get_per_client_data`.

Also, if people want to query multiple data sources (e.g. I needed to query `main_summary` for some metrics but could get away with `clients_daily` for most) then they can reuse the one `enrollments` dataframe.

Likewise if people want data from multiple conversion windows (e.g. to build a time series) they can reuse `enrollments` - though in the future we'll probably have a method separate to `get_per_client_data` to solve that need.

When people don't need flexibility then they'll be able to use mozreport to drive all this

src/mozanalysis/experiment.py
felixlawrence: Under the rug it goes!

src/mozanalysis/experiment.py
harterrt: Awesome, makes a lot of sense!

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
harterrt: Ah! I didn't know about the experiment `events` table! That makes a lot of sense.

Part of the question still holds - why not query `events` directly to identify the start date. But, I believe you effectively answered this with the `get_per_client_data` discussion. i.e. we want flexibility here and we can always have stronger defaults in mozreport. 

Thanks for walking me through this!

src/mozanalysis/experiment.py
harterrt: You can consider this my explicit buy-in, fwiw.

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
felixlawrence: > why not query `events` directly to identify the start date
Yes, this would be safer, but there is one final chicken that needs to come before that egg: when querying `events` we should specify a minimum date so that we can efficiently filter out prehistoric events.

So the current need to supply the "maximum lower bound" on relevant data could be relaxed to only require a lower bound, which would be a bit safer. Is that benefit worth losing this method's laziness? *shrugs*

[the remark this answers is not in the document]
harterrt: Strong point. I agree, the current implementation makes a lot more sense.

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,508 @@
...
+                - branch (str)
+        """
+        if study_type == 'pref_flip':
+            enrollments = self._get_enrollments_view_normandy(spark)
+
+        elif study_type == 'addon':
+            enrollments = self._get_enrollments_view_addon(spark, self.addon_version)
+
+        # elif study_type == 'glean':
+        #     raise NotImplementedError
```
tdsmith: Is the `df` argument to `get_per_client_data` the other thing we think we'd need to change to make something Gleanlike work?

src/mozanalysis/stats.py
```diff
@@ -1,66 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
tdsmith: Is it premature to remove stats.py?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,508 @@
...
+    def get_per_client_data(
+        self, enrollments, df, metric_list, last_date_full_data, conv_window_start_days,
+        conv_window_length_days, keep_client_id=False
+    ):
+        """Return a DataFrame containing per-client metric values.
+
+        Args:
+            enrollments: A spark DataFrame of enrollments, like the one
+                returned by `self.get_enrollments()`.
+            df: A spark DataFrame containing the data needed to calculate
```
tdsmith: I'd love a more informative public name for this than `df`. `ping_data`? `metric_source`? `measurements`?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,508 @@
...
+
+            This format - the schema plus there being one row per
+            enrolled client, regardless of whether the client has data
+            in `df` - was agreed upon by the DS team, and is the
+            standard format for queried experimental data.
+        """
+        req_dates_of_data = conv_window_start_days + conv_window_length_days
+
+        if self.num_dates_enrollment is None:
+            self._print_enrollment_window(last_date_full_data, req_dates_of_data)
```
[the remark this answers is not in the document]
felixlawrence: This isn't a case of "I did print statement debugging then forgot to remove the print statement".

The situation is that this function asks for inputs that are in an easy-to-specify-but-easy-to-misinterpret format, and this prints those inputs' consequences in a harder-to-misinterpret format. I figured it was worth putting this information in front of everyone.

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,508 @@
...
+
+        Args:
+            enrollments: A spark DataFrame of enrollments, like the one
+                returned by `self.get_enrollments()`.
+            df: A spark DataFrame containing the data needed to calculate
+                the metrics. Could be `main_summary` or `clients_daily`.
+                _Don't_ use `experiments`; as of 2019/04/02 it drops data
+                collected after people self-unenroll, so unenrolling users
+                will appear to churn. Must have at least the following
+                columns:
```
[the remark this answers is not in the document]
felixlawrence: Spark should immediately give an `AttributeError` if any of these columns are missing. Is that not clear enough?
```
> cd = spark.table('clients_daily')
> cd.does_not_exist
AttributeError: 'DataFrame' object has no attribute 'does_not_exist'
```

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,508 @@
...
+        `telemetry_shield_study_parquet`. Once this issue is resolved,
+        we will probably start using normandy events for all desktop
+        studies.
+        Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1536644
+
+        Args:
+            spark: The spark context.
+            study_type (str): One of the following strings:
+                - 'pref_flip'
+                - 'addon'
```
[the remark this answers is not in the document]
felixlawrence: If you have some wild experiment where this method doesn't meet your needs, you're welcome to construct your own `enrollments` dataframe and supply that to `get_per_client_data()` - I think that writing the callable is not much less effort than constructing the final dataframe? Though if patterns emerge among unsupported cases then we should try to support them.

tests/test_experiment.py
```diff
@@ -0,0 +1,361 @@
...
+
+    # When we have complete data for 20190114...
+    the_fourteenth = '20190114'
+
+    # ...we have 14 dates of data for those who enrolled on the 1st
+    exp._print_enrollment_window(the_fourteenth, 14)
+
+    # We don't have 14 dates of data for the 8-day cohort:
+    with pytest.raises(AssertionError):
+        exp_8d._print_enrollment_window(the_fourteenth, 14)
```
tdsmith: hm, these validations don't feel like something that ought to be the responsibility of something named `_print` -- are these tests all redundant with the next method? If these tests don't provide additional coverage, maybe they're safe to delete

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
tdsmith: fair -- the glean comment just caught my eye and i started thinking about what we'd have to do in order to prototype something that would snap into this framework. "bring your own dataframe" is a reasonable solution.

passing in strings also means you have to monkeypatch out those methods to unit test this function, but you already did that work. :p

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
tdsmith: No, that sounds fine.

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
tdsmith: I think "you passed in a dataframe with the wrong shape" and "something in the middle of this function threw an AttributeError" is a meaningful difference but otoh it's probably hard to hit this error if you're on the happy path

src/mozanalysis/stats.py
felixlawrence: I want to make several breaking changes to `stats.py` so I don't want people to start using this version.

After this PR, the next step is to issue another PR that adds back some stats functionality (including some lines of code lifted from this original file). This second PR will also lead to a lot of discussion, but the issues there are mostly orthogonal to the issues in this PR so I'd rather have these discussions sequentially rather than simultaneously.

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
felixlawrence: OK I guess I had my head down in the weeds - `AttributeError: 'DataFrame' object has no attribute 'client_id'` would make me immediately think that `df` was missing the column `client_id`. But someone who didn't write this code may not jump to the same conclusion :P

src/mozanalysis/experiment.py
felixlawrence: IIRC Glean is going to send enrollment events promptly - I think to the `events` table. In the long term it will probably be worth making a `glean_enrollments` table in the ETL but in the short term we'd probably just filter the events table? So we'd add a method `_get_enrollments_view_glean()`, and call it from this point in `get_enrollments()`.

The implementation of `get_per_client_data` shouldn't have to change - the user would just supply Glean's equivalent to `main_summary` as `df` (which I think is the part that directly answers your question?)

tests/test_experiment.py
felixlawrence: Those validations are the responsibility of `_get_scheduled_max_enrollment_date()`. The output of said method is used:
* In `get_enrollments()` to filter the enrollments dataframe before returning it,
* In `filter_enrollments_for_conv_window()` (called in `get_per_client_data()`) to filter the enrollments dataframe before joining to `df`
* In `_print_enrollment_window()` to inform the user which date is going to be used by `filter_enrollments_for_conv_window()`.

I do think that the validation should happen at the time that this date is calculated, so this validation does belong in `_get_scheduled_max_enrollment_date()`.

w.r.t. the automated test suite, sure, this set of tests doesn't capture much of value that's not captured by `test_get_last_enrollment_date()`, so I'll delete the vast majority of the tests in `test_check_windows()` and rename it.

src/mozanalysis/experiment.py
felixlawrence: I'd love a better name too, but this is trying to be a very generic quantity: as per the docs it is "A spark DataFrame containing the data needed to calculate the metrics". This data won't always be "ping data" or "measurements" (e.g. I don't think crash reports fall into either category). Is there a terse name that conveys "data that is going to be turned into a metric"? I think `metric_source` suggests that the data is already metricified (or emetricuted?) but I want to convey that it's OK to chuck in generic tables?

tdsmith: Yeah I was reaching for suggestions. I probably assert a row of crash_summary reflects a measurement of a crash, and it's definitely a summary of a ping; &shrug;!

setup.py
```diff
@@ -20,11 +22,9 @@
     install_requires=[
         'numpy',
         'pandas',
-        'py4j',
-        'pyarrow>=0.8.0',
-        'pyspark',
         'scipy',
         'statsmodels>=0.9',
+        'pyarrow>=0.8.0',
```
teonbrooks: where is pyarrow being used? i don't see it being import anywhere

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+class Experiment(object):
+    """Get DataFrames of experiment data; store experiment metadata.
+
+    The methods here query data in a way compatible with the following
+    principles, which are important for experiment analysis:
+
+    - The population of clients in each branch must have the same
+        properties, aside from the intervention itself and its
```
teonbrooks: we might not need to do so here but we, in general, use inconsistent terms. for example, we say intervention here but treatment is used elsewhere (branch names in experiments). we should unify how we discuss these terms

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,545 @@
...
+        consequences; i.e. there must be no underlying bias in the
+        branch populations.
+    - Every client should have an equal opportunity to convert, so that
```
teonbrooks: what do we mean by convert here?

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,545 @@
...
+        start_date (str): e.g. '20190101'. First date on which enrollment
+            events were received.
+        num_dates_enrollment (int, optional): Only include this many days
+            of enrollments. If `None` then use the maximum number of days
+            as determined by the metric's conversion window and
+            `last_date_full_data`. Typically `7n+1`, e.g. `8`. The
+            factor '7' removes weekly seasonality, and the `+1` accounts
+            for the fact that enrollment typically starts a few hours
+            before UTC midnight.
+        addon_version (str): The version of the experiment addon. Some
```
teonbrooks: (str) -> (str, optional)

src/mozanalysis/experiment.py
```diff
@@ -0,0 +1,545 @@
...
+            # Now do a more thorough pass filtering out irrelevant data:
+            # TODO perf: what is a more efficient way to do this?
+            (
+                (
+                    F.unix_timestamp(data_source.submission_date_s3, 'yyyyMMdd')
+                    - F.unix_timestamp(enrollments.enrollment_date, 'yyyyMMdd')
+                ) / (24 * 60 * 60)
+            ).between(
+                conv_window_start_days,
+                conv_window_start_days + conv_window_length_days - 1
```
teonbrooks: do you exclude the last day of the conversion window?

tests/test_experiment.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def test_print_enrollment_window():
+    exp = Experiment('a-stub', '20190101')
+
+    # When we have complete data for 20190114...
+    the_fourteenth = '20190114'
+
+    # ...we have 14 dates of data for those who enrolled on the 1st
+    exp._print_enrollment_window(the_fourteenth, 14)
```
teonbrooks: this step seems to be a manual check. how about the helper return the list of dates and you can assert that it is the length you expect

teonbrooks:
just for clarification, when you say conversion window, are you referring to the study period?

setup.py
tdsmith: cf https://github.com/mozilla/mozanalysis/pull/38/files/9f79049305a165e367ad2bea912d31533bf89b8e#diff-2eeaed663bd0d25b7e608891384b7298 -- pyspark will vomit if you try to define a pandas udf without a recentish pyarrow

src/mozanalysis/experiment.py
felixlawrence: `between` is inclusive, so this should obtain data from `conv_window_length_days` consecutive submission dates, unless I stuffed it up?

I frequently fall into this trap or at the very least second-guess myself here, so I'm looking forward to hiding these choices behind an interface that accepts an easily interpretable quantity (`conv_window_length_days`), and has automated tests with examples of how it should behave.

tests/test_experiment.py
felixlawrence: Ergh, this is a vestigial test. tdsmith correctly pointed out that the previous version of `test_print_enrollment_window` was mostly redundant due to `test_get_last_enrollment_date`, so I removed most of `test_print_enrollment_window`. I left this one statement to make sure that the method doesn't raise an exception (i.e. that the `format` parameters weren't mangled?).

But yeah it's barely worth testing, and in fact the `_print_enrollment_window()` method itself has shrunk over various refactorings to the point that I don't think it should exist any more. So I'm going to roll the print statement into `_get_last_enrollment_date()` and delete `_print_enrollment_window()` and `test_print_enrollment_window()`.

src/mozanalysis/experiment.py
felixlawrence: I have a bias towards binary metrics, where we measure whether or not a user "converts" (does the thing of interest) during a particular period of time (the conversion window, e.g. "the first week after they enroll"). So that's what I meant, but I agree it doesn't make sense in this more general context. I'll try to think of something that fits better - have you got any ideas?

felixlawrence:
> just for clarification, when you say conversion window, are you referring to the study period?

A metric (e.g. "how many active_hours did this client have?") needs to be queried with respect to a conversion window (e.g. "in their second week after enrollment").

If we want to measure users' active hours in their first week after enrolling, and also measure users' active hours in their second week after enrolling, then those are two different conversion windows.

I think the "study period" is defined with respect to the calendar and defines the time that the experiment needs to be active - enrolling and/or observing users.

A "conversion window" is defined with respect to users' moments of enrollment and defines the time window for which we're analysing data for a specific calculation

The name "conversion window" makes a lot of sense if we're considering binary metrics - "did this user meet this criterion (convert) or not?" As per [this comment](https://github.com/mozilla/mozanalysis/pull/38#discussion_r280154077) it's imperfect for our needs and suggestions are welcome.

src/mozanalysis/experiment.py
felixlawrence: I whipped up a [gist](https://gist.github.com/felixlawrence/5d67e7a5beafae32a89ed2eada794b98) and posted it to the team slack. I think this conversation is outside the scope of this PR. Are you happy enough with the language here that you'd be OK with this being merged, with the understanding that we can change it later if we decide on different language?

felixlawrence: For posterity, the gist became a [gdoc](https://docs.google.com/document/d/16SGUwosOfwY6-eaxPj1ZyoWjJU0BZLFroQVPSnUdGoI/).

ethompso28: can you put this doc in our Team google drive? eventually this should be attached to our Onboarding doc, but great to see this start

felixlawrence: It's [already there](https://drive.google.com/drive/folders/1LIVi1MQ58VLCMfkRy34pf-j1GS60HbFB)

src/mozanalysis/experiment.py
saptarshiguha: this looks good to me. I'm curious however, where does data cleaning(trimming/capping/cleaning) happen? is this something the analyst does before this spark dataframe of experiment data is returned(i.e so that the input to this function will be already cleaned)

src/mozanalysis/experiment.py
teonbrooks: I think that if it is inclusive then it should be `conv_window_start_days + conv_window_length_days` otherwise it is clipping unless I am missing something.

felixlawrence: If you set `conv_window_length_days = 1`, then how many dates of data do you expect to be returned? Would you expect data collected from one date or two? I would expect one only, and that's what I think this code does. Note that we're comparing dates, not real timestamps.

There are some examples that test this join [here](https://github.com/felixlawrence/mozanalysis/blob/revamp-and-add-query-code/tests/test_experiment.py#L308) though I'll admit that this is the one of the most convoluted parts of the test suite because it's testing something in the middle of a function

src/mozanalysis/experiment.py
felixlawrence: If you're tidying up individual rows of the data source (e.g. individual pings) then you'd need to do that before passing it to `get_per_client_data`. If you're capping or filtering outlier clients based on their values for a metric, then you'd do that with the output of `get_per_client_data`.

Does that sound reasonable and cover your usecases, or is there one I missed?

src/mozanalysis/experiment.py
teonbrooks: ah, sorry, I get it now. I was thinking in python indexing terms instead of the dates

felixlawrence: This is precisely the kind of thing that I commonly stuff up, so I'm grateful for the extra eyes :)

saptarshiguha:
All sounds good to me

src/mozanalysis/experiment.py
rjweiss: This addresses improving analysis efficiency for data science, so r+ from me.  I'll defer to @ethompso28 on whether any further review is needed (e.g. from data eng).

felixlawrence:
Just to confirm following this morning's mozanalysis meeting - I believe @sunahsuh wanted to review this PR and said she would likely do so by the end of this week - please correct this if I'm wrong.

teonbrooks:
not to be addressed in this PR but a broader question about our burgeoning codebase. do we want to follow particular style guide or best practices like numpy docstring approach, variable name styling, etc.? I'm happy to follow up with style pr

sunahsuh:
> Just to confirm following this morning's mozanalysis meeting - I believe @sunahsuh wanted to review this PR and said she would likely do so by the end of this week - please correct this if I'm wrong.

Ah, I re-read Rebecca's comment and I see she meant that she wanted @ethompso28 to weigh in on whether a data eng review was necessary, not that data eng _should_ necessarily review :)

felixlawrence:
@teonbrooks:
I question the size to which this codebase will burgeon. The package has a fairly limited scope (providing tools to query Mozilla experiment data and running statistical processing on it).

It makes sense to jump through hoops to make the package easier for DSes to use (even when their use-case deviates from the happy path). I don't think it makes sense to jump through hoops to eliminate problems that would emerge at >100K lines of code and >20 active developers working on overlapping parts of the code.

If you think that a particular style guide will help with the former and doesn't add a huge amount of overhead to contributing, then it would be very welcome. And if you want to improve the naming of function arguments then the sooner the better!

@sunahsuh:
Yep, and @ethompso28 is at cultural onboarding this week and has been involved with this project for long enough that I think (trust? hope?) that if she particularly wanted data eng review then she would have expressed that by now; I don't recall her doing so.

As a compromise/to move things along while mitigating risk, I asked you whether you wanted to review it: you said "yes" - if you want to revise that answer in light of this context, then let me know and I'll merge it.

ethompso28:
Just a note, I think it's a bit late in the game to include DE on this particular PR, but given feedback from @mreid-moz @sunahsuh and the fact that our teams should work closely together, DE should at least be more formally aware of our development moving forward, which means for future PRs it would be good to add @sunahsuh as a reviewer.

felixlawrence:
@sunahsuh on Slack:
> please go ahead and merge

PR #39, opened 2019-05-24
Work around bug in Spark joining.

felixlawrence:
If you start with a DataFrame, aggregate over it, then join from the aggregation to the original DataFrame, then Spark can raise an AnalysisException, complaining about duplicated names even if you've been careful to rename them. Specifically, this happened with 'submission_date_s3'.

Since it _actually is a unique name_, we can use `F.col('submission_date_s3')` instead of `data_source.submission_date_s3` to work around this problem. If someone supplies an `enrollments` table that _does_ have a `submission_date_s3` column then Spark justifiably raises a meaningful error.

Added the fix, a short comment explaining it with a reference to the bug[1], and a test that failed before the bug was fixed.

[1] https://issues.apache.org/jira/browse/SPARK-10925

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -295,13 +295,15 @@ class Experiment(object):
             # enrollments.branch == data_source.experiments[self.experiment_slug]
 
             # Do a quick pass aiming to efficiently filter out lots of rows:
-            enrollments.enrollment_date <= data_source.submission_date_s3,
+            # Use F.col() to avoid a bug in spark when `enrollments` is built
+            # from `data_source` (SPARK-10925)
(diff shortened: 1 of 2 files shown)


felixlawrence:
I considered checking that directly and raising a `ValueError` if it did, but:
- The happy path is to use `enrollments` from `get_enrollments()`
- `enrollments` already needs an `enrollment_date` column - it's unlikely that someone would provide that without replacing `submission_date_s3`
- I checked what happens when `enrollments` has both an `enrollment_date` column and a `submission_date` column - as alluded to in the commit message, Spark warns about `'submission_date_s3'` being an ambiguous name. I think that's enough of a hint for an adventurer who built their own `enrollments` dataframe :stuck_out_tongue:

PR #42, opened 2019-05-24
Build/test cleanups

tdsmith:
Clean up some build and test config.

Importantly, this makes docs warnings errors again, so this should fail until we patch that up.


felixlawrence:
I don't understand this area of packaging well enough to know if we're losing something useful; it would be more useful to get an engineer to review it?

README.md
```diff
@@ -16,6 +16,12 @@ pip install mozanalysis
 
 ## Testing locally
 
+### with Tox
+
+Install tox into your global Python environment and run `tox`.
```
felixlawrence: Might be friendly to point out that while they're hacking, people can run a single file of tests by something like:
```
tox -e py37 tests/stats/test_bootstrap.py
```

tdsmith:
> I don't understand this area of packaging well enough to know if we're losing something useful; it would be more useful to get an engineer to review it?

If it hasn't broken your workflow, we haven't. :) They're in git history if we decide we need them.

README.md
tdsmith: Added a couple examples.

PR #45, opened 2019-06-24
Use attrs for Experiment

tdsmith:
Use attrs for the experiment class.
This gives us more robust immutability, a free repr, and avoids some
boilerplate.


tdsmith:
https://github.com/mozilla/mozanalysis/pull/38#discussion_r279514567 strikes again :p

PR #48, opened 2019-07-29
Refactor time limits code.

felixlawrence:
Our next priorities are to introduce a metrics library and support for querying time series of metrics data. Both of these things add complexity to the queries that get sent to Spark, and the code that builds them. Here we lay some groundwork for those features by disentangling some logic from the `Experiment` class.

The main change here is the introduction of the `TimeLimits` class: an instance of TimeLimits holds all the temporal information used when querying data for an `Experiment`. It contains the start and end of the enrollment period, the details about the analysis window, and the relevant date range for the data source.

Previously, we tried to hold minimal state and compute quantities on the fly. Instead, `TimeLimits` precomputes and stores all these variables, which is conceptually easier (particularly when we start specifying a different set of variables when working with time series). We ensure self-consistency by providing a constructor with a near-minimal set of arguments, and by using validators.

This commit is not intended to be merged to master in its current state: it offers no new features or bug fixes. It is intended to form a base for future work on time series and metrics.

This commit includes a little cruft too: there is a bunch of commented-out WIP/prototype code to handle time series and a metrics library. I figured it was easier to include it here, commented out, than to maintain a complicated tree of feature branches that need to be rebased against one another. It also provides the reviewer some hints of my future intentions.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -2,9 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import attr
-from pyspark.sql import functions as F
+# from functools import reduce
+from pyspark.sql import Column, functions as F
 
 from mozanalysis.utils import add_days
+# from mozanalysis.metrics import process_metrics
 
 
 @attr.s(frozen=True, slots=True)
@@ -144,6 +146,10 @@ class Experiment(object):
                 set ``num_dates_enrollment`` then do not set this; at best
                 it would be redundant, at worst it's contradictory.
 
+            debug_dupes (bool, optional): Include a column ``num_events``
+                giving the number of enrollment events associated with
+                the ``client_id`` and ``branch``.
+
         Returns:
             A Spark DataFrame of enrollment data. One row per
             enrollment. Columns:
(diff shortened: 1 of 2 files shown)


felixlawrence:
My plan of attack is to get this PR approved then issue more PRs against this branch then only merge to master once there are useful features and/or no egregious commented-out placeholder code.

src/mozanalysis/experiment.py
```diff
@@ -285,21 +294,131 @@ def get_per_client_data(
...
+    def _get_join_conditions(
+        self, enrollments, data_source, time_limits
+    ):
+        """Return a list of join conditions.
+
+        In ``_get_results_for_one_data_source``, we join ``enrollments``
+        to ``data_source``. This method returns the list of join
+        conditions.
+
+        The join conditions are:
```
felixlawrence: Is this documentation garbage? I'm just restating the code and the comments that are already in the code... Shall I just remove it?

felixlawrence:
LMK if you would like me to separate out into a later PR the other changes to `Experiment` (refactoring `_get_per_client_data()` into `_get_results_for_one_data_source()`, `_get_join_conditions()`, etc). It would make the diff easier to read.

felixlawrence:
Breaking this down into two PRs, for ease of review. See #49 for the first of these.

PR #50, opened 2019-08-07
Further refactoring of Experiment class

felixlawrence:
Do more groundwork in preparation for supporting querying of time series and a metrics library.

Specifically, spin off the construction of the join conditions into a separate method so that `get_per_client_data()` fits on one screen and is easier to reason about.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -417,7 +417,8 @@ class Experiment(object):
             tssp.payload.addon_version.alias('addon_version'),
         )
 
-    def _process_enrollments(self, enrollments, time_limits):
+    @staticmethod
+    def _process_enrollments(enrollments, time_limits):
         """Return ``enrollments``, filtered to the relevant dates.
 
         Ignore enrollments that were received after the enrollment
@@ -428,7 +429,8 @@ class Experiment(object):
             enrollments.enrollment_date <= time_limits.last_enrollment_date
         )
 
-    def _process_data_source(self, data_source, time_limits):
+    @staticmethod
+    def _process_data_source(data_source, time_limits):
         """Return ``data_source``, filtered to the relevant dates.
 
         Ignore data before the analysis window of the first enrollment,
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -297,13 +297,9 @@ class Experiment(object):
             analysis_length_days, self.num_dates_enrollment
         )
 
-        enrollments = self._process_enrollments(
-            enrollments, time_limits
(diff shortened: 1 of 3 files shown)


src/mozanalysis/experiment.py
```diff
@@ -309,6 +309,33 @@ def get_per_client_data(
...
+            return res
+        else:
+            return res.drop(enrollments.client_id)
+
+    def _get_join_conditions(self, enrollments, data_source, time_limits):
+        """Return a list of join conditions.
+
+        In ``_get_results_for_one_data_source``, we join ``enrollments``
+        to ``data_source``. This method returns the list of join
+        conditions.
```
tdsmith: docs nit: returns a list of boolean `Column`s representing join conditions between the enrollments `DataFrame` and data_source. When used with `DataFrame.join` to perform a left-join of the data_source onto the enrollments, these conditions produce a `DataFrame` containing the rows from data_source for enrolled users, from the time of their enrollment until the end of the experiment.

src/mozanalysis/experiment.py
```diff
@@ -427,7 +423,7 @@ def _process_enrollments(enrollments, time_limits):
         """
         return enrollments.filter(
             enrollments.enrollment_date <= time_limits.last_enrollment_date
-        )
+        ).alias('enrollments')
```
tdsmith: can you add this to the docstring and/or tests as a postcondition?

src/mozanalysis/experiment.py
```diff
@@ -446,7 +442,7 @@ def _process_data_source(data_source, time_limits):
                 time_limits.first_date_data_required,
                 time_limits.last_date_data_required
             )
-        )
+        ).alias('data_source')
```
tdsmith: can you add this to the docstring and/or tests as a postcondition?

tdsmith:
(thank you thank you thank you for the small commits!)

src/mozanalysis/experiment.py
felixlawrence: "from the time of their enrollment until the end of the experiment" is not quite correct in general - it only contains the rows received during the analysis window, which is typically a subset of that time. But I've incorporated the gist of your message:
```lang=py
        """Return a list of join conditions.

        Returns a list of boolean ``Column``s representing join
        conditions between the ``enrollments`` ``DataFrame`` and
        ``data_source``.

        In ``_get_results_for_one_data_source``, we left join
        ``enrollments`` to ``data_source`` using these join conditions
        to produce a ``DataFrame`` containing the rows from
        ``data_source`` for enrolled clients that were submitted during
        the analysis window.
        """
```

PR #54, opened 2019-12-07
Move to bigquery

felixlawrence:


felixlawrence:
ugh i can't make it a draft after creating it?

src/mozanalysis/metrics/__init__.py
```diff
@@ -2,345 +2,118 @@
...
 
-def agg_any(col):
-    """Return the bitwise OR, as an int, with 0-filled nulls.
-
-    Args:
-        col (Column): A Spark Column of bools
-    """
-    return F.coalesce(F.max(col).astype('int'), F.lit(0))
+def agg_any(select_expr):
+    """Return the bitwise OR, as an int, with 0-filled nulls."""
```
tdsmith: s/bitwise/logical/

(Why do we prefer an integer here?)

src/mozanalysis/metrics/desktop.py
```diff
@@ -2,95 +2,84 @@
...
+    select_expr=agg_sum('organic')
+)
+
+unenroll = Metric(
+    name='unenroll',
+    data_source=events,
+    select_expr=agg_any("""
+            event_category = 'normandy'
+            AND event_method = 'unenroll'
+            AND event_string_value = '{experiment_slug}'
```
tdsmith: What are the rules for templating?

src/mozanalysis/metrics/__init__.py
felixlawrence: Whoa! This is Spark code. Can you tell that we don't unit test our metrics?

At least colab's `!pip install` can install branches from github, so I'll try out these metrics before merging (after fixing this code)

felixlawrence:
Fair points re: RowIterators. I agree that we're mainly going to work with complete DataFrames. Perhaps there are three use-cases, in exponentially decaying order of commonality:

1. Non-lazy: just get complete DataFrames, and dicts of complete DataFrames
2. Long time series with lots of experiment subjects: get some custom return type (as you suggest) that returns one DataFrame at a time. It could call `.query()` on the fly so that we don't re-use bigquery result iterators.
3. Crazy big experiments with enormous datasets: you can write custom queries on the full experiment results table.

> What do we do if we want to compute the results at time A but work with them at some later time B?

Do we really want to do that? Introducing async stuff comes with an extra learning curve; do we have a use-case to justify it?

Agree with the other comments. "BigqueryStuff" was intentionally unprofessionally named, because it's an unprofessional object (it's a mashup of a client and some configuration); I'll get over my little protest

> Are performance histograms in-spec for mozanalysis?... the mozanalysis thing to do

I would like them to be in-spec. IMO the mozanalysis thing to do would be to define each metric explicitly (without using UDFs or python functions), then to get frustrated with the repetition and abstract it away using python functions that do string manipulation - which is the same end state as you proposed (though you seemed to bypass the "get frustrated" stage)

> ironically, I'm actually losing some type safety here.
...

src/mozanalysis/metrics/__init__.py
tdsmith: I think it's correctly operating elementwise -- it's just not doing bitwise operations (e.g. the result of "`agg_any([1, 64])`" will be 1, not 65)

src/mozanalysis/metrics/desktop.py
felixlawrence: `{experiment_slug}` is provided; IIRC nothing else is.

If you get stuck and need other templating, then you can define your metrics on the fly, scratching your own itch. If this is a common occurrence then we can add extra templating as required, safe in the knowledge that we're not blocking anyone in the meantime

tdsmith:
>> What do we do if we want to compute the results at time A but work with them at some later time B?
>
> Do we really want to do that? Introducing async stuff comes with an extra learning curve; do we have a use-case to justify it?

I don't necessarily think we want to use async patterns anywhere; I just mean that after we run one of these queries, the destination tables will be hanging around somewhere, and wouldn't it be nice if they were already ready to use when we booted up the ol' VM tomorrow. I guess that's what the exception path in run_query does.

felixlawrence:
> I don't necessarily think we want to use async patterns anywhere; I just mean that after we run one of these queries, the destination tables will be hanging around somewhere, and wouldn't it be nice if they were already ready to use when we booted up the ol' VM tomorrow. I guess that's what the exception path in run_query does.

Yep - if you come back the next day and re-run your notebook, it will just pull it out of yesterday's destination table :)

src/mozanalysis/metrics/__init__.py
felixlawrence: > Whoa! This is Spark code

:man_facepalming: This is _removing_ Spark code. I was not in good shape yesterday!

> s/bitwise/logical/

[lol](https://github.com/mozilla/mozanalysis/pull/26/files/a13e18c20d7021daf6e35a90ca9e19240472d641#diff-71962d4e0d4376ffe13d532605fbfbf8)

> Why do we prefer an integer here

Good question; this is a hangover from our spark datasets. I'll have to check whether we can now rely on things being boolean.

src/mozanalysis/bq_util.py
```diff
@@ -0,0 +1,54 @@
...
+    of_good_character_but_possibly_verbose = re.sub(r'[^a-zA-Z_0-9]', '_', table_name)
+
+    if len(of_good_character_but_possibly_verbose) <= 1024:
+        return of_good_character_but_possibly_verbose
+
+    return of_good_character_but_possibly_verbose[:500] + '___' \
+        + of_good_character_but_possibly_verbose[-500:]
+
+
+class BigqueryStuff(object):
```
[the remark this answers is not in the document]
felixlawrence: Let's strip out python 2 support in a different PR. I've addressed your other comments.

src/mozanalysis/metrics/__init__.py
felixlawrence: I've fixed this; there is one remaining question in my mind: do we stick with "any/all" naming or do we switch to "or/and" naming (i.e. rename it `agg_or()`).

IIRC, in spark land, "any/all" was a no-brainer because it was consistent with spark. But we're not there any more.

I'm leaning towards keeping `agg_any()` because I think it's clearer in the context in which `agg_any()` is being used:
```
unenroll = Metric(
    name='unenroll',
    data_source=events,
    select_expr=agg_any("""
            event_category = 'normandy'
            AND event_method = 'unenroll'
            AND event_string_value = '{experiment_slug}'
        """)
)
```

docs/guide.rst
```diff
@@ -7,39 +7,47 @@ Basic experiment: get the data & crunch the stats
...
     import mozanalysis.bayesian_stats.binary as mabsbin
     from mozanalysis.experiment import Experiment
+    from mozanalysis.bq import BigQueryContext
+
+
+And get a BigQuery context (a client, and some config)::
 
-For querying data, the general approach of :mod:`mozanalysis` is to start by obtaining a list of who was enrolled in what branch, when. This list is the ``enrollments`` Spark DataFrame; it has one row per client. Then we try to quantify what happened to each client: for a given analysis window (a specified period of time defined with respect to the client's experiment results.date), we seek to obtain a value for each client for each metric. So we end up with a results (pandas) DataFrame with one row per client and one column per metric.
+    bq_context = BigQueryContext(
+        dataset_id='{your_dataset_id}',  # e.g. mine's 'flawrence'
```
tdsmith: I think this is clearer without the {}'s.

docs/guide.rst
```diff
@@ -7,39 +7,47 @@ Basic experiment: get the data & crunch the stats
...
+    )
+
+If you're not part of ``moz-fx-data-bq-data-science``, then you'll also need to pass a ``project_id`` argument when initializing ``BigQueryContext()``.
```
tdsmith: What does it mean to be part of moz-fx-data-bq-data-science? There's some kind of persistent default project somewhere, right?

tdsmith: Oh, I see; "to query using a project other than..."

src/mozanalysis/metrics/desktop.py
tdsmith: Okay; please document that on the Metric class.

src/mozanalysis/bq.py
```diff
@@ -0,0 +1,63 @@
...
+
+    .. _BigQuery dataset id: https://cloud.google.com/bigquery/docs/datasets
+    """
+    def __init__(self, dataset_id, project_id='moz-fx-data-bq-data-science'):
+        self.dataset_id = dataset_id
+        self.project_id = project_id
+        self.client = bigquery.Client(project=project_id)
+
+    def run_query(self, sql, results_table=None):
+        """Run a query and return the result.
```
tdsmith: Can you document the return type?

tdsmith:
> In Python 3, it turns out that the `hash()` of a string is consistent within an interpreter session, but inconsistent between sessions.

Oh yeah, hash randomization is a security thing; this is to avoid letting users submit input that predictably grows pathological data structures that DoS the interpreter. Cats can have a little cryptographic hashing instead. :+1:

src/mozanalysis/experiment.py
```diff
@@ -330,299 +228,167 @@ def get_time_series_data(
...
-        """Like ``get_time_series_data()`` but with Spark DataFrames.
-
-        See docs for ``get_time_series_data()``.
-
-        Sometimes the results of ``get_time_series_data()`` do not fit
-        in memory. This method returns a dict of Spark DataFrames,
-        instead of a dict of pandas DataFrames. This allows for
-        subsequent processing to be done in Spark, or for smaller
-        portions of data to be loaded into memory.
+            Also returns a ``google.cloud.bigquery.table.RowIterator``
```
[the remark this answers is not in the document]
felixlawrence: A vestigial lie...

src/mozanalysis/experiment.py
```diff
@@ -849,3 +615,89 @@ def _validate_start(self, attribute, value):
...
+        """
+        return """
+            SELECT * EXCEPT (client_id, analysis_window_start, analysis_window_end)
+            FROM `{project_id}.{dataset_id}.{full_table_name}`
+            WHERE analysis_window_start = {aws}
+            AND analysis_window_end = {awe}
+        """.format(
+            project_id=bq_context.project_id,
+            dataset_id=bq_context.dataset_id,
+            full_table_name=self.full_res_table_name,
```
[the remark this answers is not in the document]
felixlawrence: Solving this was less gross than I feared - we don't need to solve this problem at the caching layer (i.e. inside `BigQueryContext.run_query()`) - just for this class's query-building method.

I'd appreciate it if you took another look to see if this solution is sufficiently tasteful.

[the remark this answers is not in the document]
tdsmith: Sure, this works.

PR #55, opened 2019-12-30
 Make bootstrapping work without Spark.

felixlawrence:
This makes a breaking change to the bootstrap function signatures: arguments are re-ordered so that `sc` becomes optional.

Mozanalysis will now be frequently run without a Spark context. We'll still need to bootstrap.

Now if no Spark context is supplied, non-parallel implementations of the bootstrap will be run.

### DIFF
--- a/src/mozanalysis/bayesian_stats/bayesian_bootstrap.py
+++ b/src/mozanalysis/bayesian_stats/bayesian_bootstrap.py
@@ -97,15 +97,15 @@ def make_bb_quantile_closure(quantiles):
 
 
 def compare_branches(
-    sc, df, col_label, ref_branch_label='control', stat_fn=bb_mean,
+    df, col_label, ref_branch_label='control', stat_fn=bb_mean,
     num_samples=10000, threshold_quantile=None,
     individual_summary_quantiles=mabs.DEFAULT_QUANTILES,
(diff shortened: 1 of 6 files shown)


tdsmith:
Yeah, this looks straightforward. It would be useful to optionally distribute these across a process pool in the local-machine case in order to exploit a big n-core cloud box but we can do that later (though aiui if we're in processland, we're not sharing memory, which means each independent process has to have its own copy of the data we're bootstrapping, which maybe sucks).

For tests we can stay DRYer with parameterization. Writing a `spark_context_or_none` fixture using the strategy at https://docs.pytest.org/en/latest/example/parametrize.html#deferring-the-setup-of-parametrized-resources seems clean.

felixlawrence:
Fixture parametrisation is done.

Yes, there are two low-hanging optimisations here:
* locally parallelising things
* partially unrolling the loop, to take at least e.g. 1e6 samples each iteration, potentially doing multiple batches simultaneously

I considered them to be out of scope for this PR (which aims to change the status quo from unusable in colab to usable in colab), and of unknown usefulness. E.g. I'm not sure how much of a productivity win it would be, particularly given that in the foreseeable future these functions may mostly run in batch jobs, with parallelism over experiments rather than within experiment scripts.

PR #56, opened 2020-02-11
Add metric for connecting a firefox account

felixlawrence:
In order to add 'connect_fxa' as a metric, I had to un-break the join to the `events` table (the changes in `metrics/__init__.py`).

I applied this quickly to [an existing onboarding experiment](https://colab.research.google.com/drive/1fQQVj7yB3x9NHIRxI371_Pk0VEGxXV9W) and the metric result was nonzero and plausible (to me).

@tdsmith: can you please review the changes to `__init__.py` - it would be particularly appreciated if you can think of a better name than 'map_struct'.

@irrationalagent: could you please check that the metric is defined correctly and named sensibly? My [application notebook](https://colab.research.google.com/drive/1fQQVj7yB3x9NHIRxI371_Pk0VEGxXV9W) has the full generated query, if you want to see how the `Metric` object is used to form a concrete SQL query.

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -8,7 +8,7 @@ import attr
 class DataSource(object):
     name = attr.ib(validator=attr.validators.instance_of(str))
     from_expr = attr.ib(validator=attr.validators.instance_of(str))
-    experiments_column_type = attr.ib(default='desktop_main_ping', type=str)
+    experiments_column_type = attr.ib(default='simple', type=str)
     client_id_column = attr.ib(default='client_id', type=str)
     submission_date_column = attr.ib(default='submission_date', type=str)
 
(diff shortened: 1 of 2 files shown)


src/mozanalysis/metrics/__init__.py
```diff
@@ -90,6 +98,24 @@ def get_sanity_metrics(self, experiment_slug):
...
+                    select_expr=agg_any("""`moz-fx-data-shared-prod.udf.get_key`(
+                ds.experiments, '{experiment_slug}'
+            ).branch != e.branch"""),
+                ),
+                Metric(
+                    name=self.name + '_has_non_enrolled_data',
+                    data_source=self,
+                    select_expr=agg_any("""`moz-fx-data-shared-prod.udf.get_key`(
+                ds.experiments, '{experiment_slug}'
+            ).branch IS NULL""".format(experiment_slug=experiment_slug))
```
tdsmith: This indentation is a little hard to follow but I assume it does something nice looking on the output side

PR #59, opened 2020-02-21
Add info to guide about creating datasets.

felixlawrence:


tdsmith:
@benmiroglio Would this have been clearer?

benmiroglio:
I'd just add a note about "this is the space where the analysis below will write tables for future use." BQ's terminology, calling the space a "dataset" is really what made this confusing at first (even being well aware of the terminology!).

felixlawrence:
@benmiroglio Do these changes fix that?

benmiroglio:
@felixlawrence Yes, that is great. Thanks for the quick change!

felixlawrence:
No worries! tdsmith sneakily reassigned you as the reviewer - could you please hit the button so I can merge?

PR #61, opened 2020-02-25
Add documentation for DataSource

felixlawrence:


src/mozanalysis/metrics/__init__.py
```diff
@@ -6,6 +6,29 @@
...
+            table name. Sometimes a subquery.
+        experiments_column_type (str or None): Info about the schema
+            of the table or view:
+
+            * 'simple': There is an ``experiments`` column, which is an
+              (experiment_slug:str -> branch_name:str) map.
+            * 'native': There is an ``experiments`` column, which is an
+              (experiment_slug:str -> struct) map, where the struct
+              contains a ``branch`` field, which is the branch as a
+              string.
```
tdsmith:
```suggestion
              string.
            * None: There is no ``experiments`` column.
```

tdsmith: I'm not sure what happens in this case, but it looks supported!

PR #62, opened 2020-02-27
Add main and crash ping data sources

tdsmith:
... and a histogram-mean aggregation helper.

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -170,3 +170,12 @@ def agg_sum(select_expr):
 def agg_any(select_expr):
     """Return the logical OR, with FALSE-filled nulls."""
     return "COALESCE(LOGICAL_OR({}), FALSE)".format(select_expr)
+
+
+def agg_histogram_mean(select_expr):
+    """Produces an expression for the mean of an unparsed histogram."""
+    return dedent(f"""\
+        SAFE_DIVIDE(
+            SUM(CAST(JSON_EXTRACT_SCALAR({select_expr}, "$.sum") AS int64)),
+            SUM((SELECT SUM(value) FROM UNNEST(`moz-fx-data-shared-prod`.udf.json_extract_histogram({select_expr}).values)))
+        )""")
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -1,6 +1,8 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from textwrap import dedent
+
 import attr
 
 
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -1,8 +1,6 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
-from textwrap import dedent
-
 import attr
 
 
@@ -176,8 +174,7 @@ def agg_any(select_expr):
 
 def agg_histogram_mean(select_expr):
     """Produces an expression for the mean of an unparsed histogram."""
-    return dedent(f"""\
-        SAFE_DIVIDE(
-            SUM(CAST(JSON_EXTRACT_SCALAR({select_expr}, "$.sum") AS int64)),
-            SUM((SELECT SUM(value) FROM UNNEST(`moz-fx-data-shared-prod`.udf.json_extract_histogram({select_expr}).values)))
-        )""")  # noqa
+    return f"""SAFE_DIVIDE(
+                SUM(CAST(JSON_EXTRACT_SCALAR({select_expr}, "$.sum") AS int64)),
+                SUM((SELECT SUM(value) FROM UNNEST(`moz-fx-data-shared-prod`.udf.json_extract_histogram({select_expr}).values)))
+            )"""  # noqa
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -180,4 +180,4 @@ def agg_histogram_mean(select_expr):
         SAFE_DIVIDE(
             SUM(CAST(JSON_EXTRACT_SCALAR({select_expr}, "$.sum") AS int64)),
             SUM((SELECT SUM(value) FROM UNNEST(`moz-fx-data-shared-prod`.udf.json_extract_histogram({select_expr}).values)))
-        )""")
+        )""")  # noqa
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from textwrap import dedent
+
 from mozanalysis.metrics import Metric, DataSource, agg_sum, agg_any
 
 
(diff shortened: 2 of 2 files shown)


src/mozanalysis/metrics/desktop.py
```diff
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from textwrap import dedent
```
felixlawrence: If you're trying to be tidy about indentation, have you checked what the generated query looks like?

The existing codebase _tries_ to have sensible and consistent indentation in the finally-built query, though this is a bit of a house of cards. If you care about indentation, then please match or improve the pre-existing pattern?

src/mozanalysis/metrics/__init__.py
```diff
@@ -170,3 +170,12 @@ def agg_sum(select_expr):
...
     return "COALESCE(LOGICAL_OR({}), FALSE)".format(select_expr)
+
+
+def agg_histogram_mean(select_expr):
+    """Produces an expression for the mean of an unparsed histogram."""
+    return dedent(f"""\
+        SAFE_DIVIDE(
+            SUM(CAST(JSON_EXTRACT_SCALAR({select_expr}, "$.sum") AS int64)),
+            SUM((SELECT SUM(value) FROM UNNEST(`moz-fx-data-shared-prod`.udf.json_extract_histogram({select_expr}).values)))
+        )""")
```
felixlawrence: This will sometimes return `NULL`, won't it? Is there a way to analyse the quantity that you want to analyse that doesn't involve dropping nulls before applying any stats? E.g. counting the fraction of users above some threshold?

It's hard to rule out the possibility that dropping nulls is biasing the results in favour of one branch - and even making that check in the first place is an extra step that's easily forgotten.

So, if it's possible to achieve your analysis goals in a way that safely handles nulls without adding any footguns, then that would be preferable. If not, then we'll have to continue to live dangerously.

src/mozanalysis/metrics/desktop.py
```diff
@@ -27,6 +29,32 @@
     experiments_column_type='native',
 )
 
+main = DataSource(
+    name='main',
+    from_expr=dedent("""\
+        (
+            SELECT
+                *,
+                DATE(submission_timestamp) AS submission_date,
```
felixlawrence: Now that timestamps are more readily available (?) we should consider whether we switch our analysis window logic from dates to timestamps - which would make life simpler in several ways...

src/mozanalysis/metrics/desktop.py
```diff
@@ -27,6 +29,32 @@
     experiments_column_type='native',
 )
 
+main = DataSource(
+    name='main',
+    from_expr=dedent("""\
+        (
+            SELECT
```
felixlawrence: In how many more places are we likely to see this pattern of `SELECT *` and turn a timestamp to a date and pull `experiments` out of the environment? Is it worthwhile supporting it in the `DataSource` class so that these `from_expr` could just become table names?

My guess right now is that this would be over-optimising, but dunno?

src/mozanalysis/metrics/desktop.py
tdsmith: I'll check. I'm agnostic about the relative importance of "nice code" and "nice query." Hopefully a human never sees the latter, but if they do, it's probably worth making it intelligible.

src/mozanalysis/metrics/__init__.py
tdsmith: > This will sometimes return NULL, won't it?

Yes.

> Is there a way to analyse the quantity that you want to analyse that doesn't involve dropping nulls before applying any stats? 

No; I think the line I want to take is that users without any e.g. DNS activity simply don't inform an analysis of DNS performance. These metrics should be applied to an analysis window spanning the entire experiment, so that users that flaked because of a poor experience are still included in the result. The user average value is a natural and useful quantity. I appreciate they're not obviously appropriate for timeseries or engagement metrics but we probably don't encode the latter as histograms very often.

src/mozanalysis/metrics/desktop.py
tdsmith: They'll never be present in clients_daily, though I think e.g. main_summary has them.

src/mozanalysis/metrics/desktop.py
tdsmith: Hm, thinking out loud: all of the pre-aggregation ping tables, including the Glean tables, but otoh there just aren't that many doctypes. The experiments column is in different places in different (e.g. Glean) tables, so we'd have to add a property to the DataSource class to specify that...

I don't really like asking consumers to code against a table spec that they can't see, so maybe it makes sense to introduce some concept of "native" and "aggregated" table types?

src/mozanalysis/metrics/desktop.py
tdsmith: All right, revised.

felixlawrence: Given we're metaprogramming SQL, "print the programatically generated query and debug it directly" is going to be a vital development tool when things go wrong adding new metrics and data sources. We want to make this as easy as possible; sensible/consistent indentation makes this slightly less awful.

Rethinking indentation everywhere is definitely on the table, but we should act intentionally.

src/mozanalysis/metrics/__init__.py
felixlawrence: > These metrics should be applied to an analysis window spanning the entire experiment, so that users that flaked because of a poor experience are still included in the result.

Is there a way we can design things such that the set of "what is straightforward to query using mozanalysis" is as close as possible to the set of "what is a valid analysis"? Specifically, if this metric only makes sense for an analysis window starting at 0, should we try to enforce that (either weakly, with a warning, or strictly, with an error)?

And/or should we force-add a sanity metric checking the NULL rates in each branch?

src/mozanalysis/metrics/desktop.py
felixlawrence: whispers softly: _we could stop using `clients_daily`_

src/mozanalysis/metrics/desktop.py
felixlawrence: > The experiments column is in different places in different (e.g. Glean) tables, so we'd have to add a property to the DataSource class to specify that...

We already have a property that does something very similar to that...

> I don't really like asking consumers to code against a table spec that they can't see, so maybe it makes sense to introduce some concept of "native" and "aggregated" table types?

I'm going to ask you in person what you mean by this.

src/mozanalysis/metrics/desktop.py
tdsmith: Well, more to the point, they're also in `search_clients_daily`. We should* use UDFs on telemetry.main instead someday, but I'd prefer to kick that can down the road a ways.

* citation needed

src/mozanalysis/metrics/desktop.py
tdsmith: captured in #68

src/mozanalysis/metrics/__init__.py
tdsmith: > Is there a way we can design things such that the set of "what is straightforward to query using mozanalysis" is as close as possible to the set of "what is a valid analysis"? Specifically, if this metric only makes sense for an analysis window starting at 0, should we try to enforce that (either weakly, with a warning, or strictly, with an error)?

That introduces a linkage between a metric and a concrete analysis window, which seems like it contradicts the design philosophy. I'm optimistic that review will catch egregious errors.

> And/or should we force-add a sanity metric checking the NULL rates in each branch?

If a metric is a Metric, I don't think there's anything we could add here. Do you mean the subsequent summary ops could report the frequency of NULLs?

felixlawrence: > That introduces a linkage between a metric and a concrete analysis window, which seems like it contradicts the design philosophy. I'm optimistic that review will catch egregious errors.

It does contradict the design philosophy, but if we have a runtime check that enforces it weakly, then I consider this _bending_ rather than breaking the philosophy. I'm less optimistic about review than you are, but you've considered my argument so I'll disagree and commit.

> If a metric is a Metric, I don't think there's anything we could add here. Do you mean the subsequent summary ops could report the frequency of NULLs?

It's a common (and in my POV dangerous) request to have metrics that return NULLs, so that the NULLs can be dropped before running stats. If we're going to accept that this evil is necessary (in some cases it probably is), then we could stick some logic somewhere in mozanalysis that checks which elements of `metrics_list` can return NULLs, and adds a binary column for each of those metrics, containing whether the column value is NULL. Yes, this is redundant information, but it's putting an important easily-forgotten quantity right in front of the user.

tdsmith: I think we agree that NULLs should be acknowledged; I just think the next aggregation level is the right place to add that column.

felixlawrence: I can imagine mozilla DSes picking and choosing which parts of the stack they want to use - do they want to use the whole dashboarding solution? Do they want to use mozanalysis manually in a notebook to get the data, aggregate and run the stats? Do they just want to use the stats functions? Do they just want to get the data then use their own stats functions?

Where should we add the NULL-checking to be safe for the most people? Above, I thought it should go into the "get the data" phase to catch the most people, but maybe it should go into the stats functions - the bootstrap functions could automatically drop nulls then also check for bias and return the extra results??

felixlawrence: ...or we add new bootstrap functions that drop the nulls and call the existing bootstrap functions as well as the sanity checks - which will have better properties w.r.t. consistent return types.

felixlawrence: Sure - something like "Add bootstrap functions that support data with NULLs"?

PR #65, opened 2020-02-28
Validate experiments_column_type parameter

tdsmith:


tdsmith:
> Is there a way to make updates more lightweight?

The tests don't really need updating for additional values. We can autogenerate the validation message, at least.

> Also, should we add another experiments_column_type for main and crash, where it lies in environment.experiments?

I argue "no"; it's the same type in a different location.

Issue #60, opened 2020-02-25
Allow users to specify destination table names


tdsmith:
I'm not sure whether this is explicitly supported now but it would be nice to support "place this back in bigquery" as an output target in addition to pandas dataframes. Are the table names exposed anywhere right now? It would be useful to be able to specify destination table names to use in place of the automatically generated content-addressable table names so that we can land them to a consistent place.

Just to toss around the idea, we could add a layer of indirection by pointing a view to the table name mozanalysis gives us, but I'm not totally sure what we gain.

PR #66, opened 2020-02-28
Promote build_query to public API

tdsmith:
Rationale: pensieve wants to be opinionated about how it runs queries (e.g. specifying destination table names, maybe not materializing results immediately) in ways that we don't need to expose to interactive users. `get_time_series_data` and `get_single_window_data` are thin wrappers around this function; instantiating a TimeLimits object is something that pensieve can do. Generating this query is how mozanalysis adds value, so let's expose the query directly.

Moots #60.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -128,7 +128,7 @@ class Experiment:
                 collected outside this analysis window.
             analysis_length_days (int): the length of the analysis window,
                 measured in days.
-            enrollments_query_type (str): Specifies the query to use to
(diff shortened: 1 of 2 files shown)


tdsmith:
fyi @scholtzan

tdsmith:
> If the goal here is "make life easier for scholtzan" then go ahead, but build_query deserves a better docstring, including something that warns away end-users.

Thanks! This is my primary objective, but, hey, it does facilitate debugging :) I wanted to make sure you didn't have a fatal objection before I pushed ahead with docs; I'll go type some out.

PR #67, opened 2020-02-28
Allow Metrics to be rebased on other DataSources

tdsmith:
Sometimes you have tables with identical schemas and need to prefer one or the other in different contexts. These could be using `telemetry_live` tables, or using a Fenix Nightly table vs a Fenix release table. This helper facilitates that.

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -160,6 +160,14 @@ class Metric(object):
     data_source = attr.ib(type=DataSource)
     select_expr = attr.ib(type=str)
 
+    def from_data_source(self, data_source: DataSource) -> "Metric":
+        """Returns an instance of the Metric drawing from a different DataSource.
+
+        This is particularly useful for Glean, where pings from e.g. Fenix
+        and Fenix Nightly are sent to different tables with identical schemas.
+        """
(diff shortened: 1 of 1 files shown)


tdsmith:
I think `import a, b, c from metrics.mobile; my_metrics = [m.from_data_source(fenix_nightly_metrics) for m in (a, b, c)]` isn't too awful for an analyst to type and remains helpfully generic.

I think the `channel=` idea assumes that's the only reason we'd want this, which, eh!

felixlawrence:
> I think `import a, b, c from metrics.mobile; my_metrics = [m.from_data_source(fenix_nightly_metrics) for m in (a, b, c)]` isn't too awful for an analyst to type and remains helpfully generic.

This moves us away from being able to specify an experiment's `Metric`s for the experiment dashboard as text configuration; we'd need to supply Python code. I'm OK with this consequence but this should be a conscious decision.

tdsmith:
Nothing but expediency requires that a declarative syntax matches the procedural syntax, though it's a useful thought exercise -- we could maybe define experiments to accept some metric transformation function that would accept a list of Metrics and return a transformed list of Metrics with different properties. We could provide a Fenix release->nightly transformation function in the box.

That would probably use this mechanism, though.

PR #70, opened 2020-03-07
Fix an error message

tdsmith:
Fix the date referenced in the TimeLimits single_analysis_window constructor if we aren't collecting data long enough.


tdsmith:
Thanks!

PR #81, opened 2020-05-04
add cfr DataSource

irrationalagent:
Adds a datasource pointing to the `messaging_system.cfr` which contains the CFR metrics events.

A couple notes: the `experiments` column on this table was recently fixed, I believe it is of type `simple` (the default for the DataSource class).

I am not adding metrics here, because there are additional variable(s) `bucket_id` and `message_id` that are required to filter down to the CFR campaign of interest (in a CFR experiment, an enrolled user may see multiple CFR campaigns, so its not possible to just pull all rows from the CFR table matching an enrolled client_id).

For example, for a recent experiment my metric definition looked like this:
```python
cfr_impression = Metric(
    name='cfr_impression',
    data_source=cfr,
    select_expr=agg_any("""event = 'IMPRESSION' 
    AND bucket_id = {}
    AND message_id = {}""".format(CFR_BUCKET_ID, CFR_MESSAGE_ID))
)
```

bucket_id and message_id will vary with experiment, making it difficult to codify them in the Metrics definition file. 

We could consider extending the `Metric` class to a `CfrMetric` class that takes bucket_id and message_id as additional properties, but IMO for now that would add a bunch more boilerplate without too much of a gain in convenience. I will be adding a template notebook for CFR analysis to dscontrib so people can copypasta example metrics from there, if they want.

### DIFF
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -56,7 +56,7 @@ cfr = DataSource(
     from_expr="""(
                 SELECT
                     *,
-                    DATE(submission_timestamp) AS submission_date                
+                    DATE(submission_timestamp) AS submission_date
                 FROM `moz-fx-data-derived-datasets`.messaging_system.cfr
             )""",
 )
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -51,6 +51,16 @@ crash = DataSource(
     experiments_column_type="native",
 )
 
+cfr = DataSource(
+    name='cfr',
+    from_expr="""(
+                SELECT
+                    *,
+                    DATE(submission_timestamp) AS submission_date                
+                FROM `moz-fx-data-derived-datasets`.messaging_system.cfr
+            )""",
+)
+
 active_hours = Metric(
     name='active_hours',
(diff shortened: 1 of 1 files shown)


tdsmith:
> the experiments column on this table was recently fixed, I believe it is of type simple (the default for the DataSource class).

It's native-typed --

https://mozilla.github.io/mozanalysis/api/metrics.html?highlight=native:

> There is an experiments column, which is an (experiment_slug:str -> struct) map, where the struct contains a branch field, which is the branch as a string.

And per https://sql.telemetry.mozilla.org/queries/70667/source they look like:

> `[{"key":"searchCohort","value":{"branch":"nov17-2"}},{"key":"rollout-doh-beta-rollout-to-all-us-desktop-users-v2-beta-74-80-bug-1613489","value":{"branch":"active"}}]`

Thanks for highlighting the metrics parameter problem; we're gonna have to deal with that in pensieve in the very near term.

irrationalagent:
> And per https://sql.telemetry.mozilla.org/queries/70667/source they look like:
> 
> > `[{"key":"searchCohort","value":{"branch":"nov17-2"}},{"key":"rollout-doh-beta-rollout-to-all-us-desktop-users-v2-beta-74-80-bug-1613489","value":{"branch":"active"}}]`

doh

PR #83, opened 2020-05-07
Update usage regularity segments to v3.

felixlawrence:
Docs for these segments are [on DTMO](https://docs.telemetry.mozilla.org/concepts/segments.html#regular-users-v3).

Here's a [colab notebook](https://colab.research.google.com/drive/1C1y3DbZ4vN1-gIh5JrIcQL8c8kZv8WoK) that demonstrates that this generates valid SQL that executes against our DB!

### DIFF
--- a/src/mozanalysis/segments/desktop.py
+++ b/src/mozanalysis/segments/desktop.py
@@ -13,64 +13,18 @@ clients_last_seen_lag_1_day = SegmentDataSource(
 )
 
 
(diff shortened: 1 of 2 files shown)


felixlawrence:
Policy question: should we be removing v2 as we add v3? I think "yes at the moment" because no-one was using it AFAIK, but in the future we probably want to keep things around for a while?

PR #84, opened 2020-05-12
Initialize Series with explicit dtype

tdsmith:
Silences a DeprecationWarning that the default dtype will change.


tdsmith:
> I guess the formatting changes in this PR are enough of a win to push this over the line.

Oh, those were accidental. Let me remove them. 😅 

Instantiating an empty Series is a little unusual and we do want to fill it with floats, so *gestures* this is fine

PR #90, opened 2020-06-16
num_dates_enrollment must be positive

### LINKED ISSUE #74: TimeLimits.for_ts throws an assertion when num_dates_enrollment = 0
I think an experiment with num_dates_enrollment = 0 is probably unreasonable, but I think the failure mode should be to emit a ValueError. Does that sound right to you, @felixlawrence ?

tdsmith:
So give a nicer error message. Fixes #74.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -623,6 +623,9 @@ class TimeLimits:
                 time_series_period
             ))
 
+        if num_dates_enrollment <= 0:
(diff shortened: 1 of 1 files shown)


tdsmith:
Thanks! I noticed this because Experimenter has experiments with no defined enrollment period, which ends up being represented as "zero dates enrollment"; pensieve has its own logic to decline to process those experiments now, but I think "all visible asserts are bugs" is a good rule, so this was on my to-do list, which I'm sorting through...

PR #91, opened 2020-06-18
normandy_events datasource for unenroll metric

scholtzan:
In pensieve, calculating the unenroll metric on a daily basis is quite expensive since currently `telemetry.events` is queried directly without any filtering. If we make use of  the `event_category` partitioning running the query only costs us a fraction.

I'm not sure if unenroll results are still correct with these changes though...

### DIFF
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -62,6 +62,16 @@ cfr = DataSource(
     experiments_column_type="native",
 )
 
+normandy_events = DataSource(
+    name='normandy_events',
+    from_expr="""(
+        SELECT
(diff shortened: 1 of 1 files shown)


felixlawrence:
Nice!

It might be worth dropping a comment into `normandy_events` explaining why using this should be preferred to `events`.

And we should consider dropping the redundant `event_category = 'normandy'` from the metric definition.

Aside:
I hope that this isn't the beginning of the end for the `Metric`/`DataSource` separation; if this kind of thing is required frequently then we should look at what we can do to expose the filtering buried in each `Metric`, in such a way that BQ can use the filtering to do query optimisation :/

PR #97, opened 2020-09-02
Declare a dtype at Series creation

tdsmith:


tdsmith:
If they change it we'll definitely get another round of DeprecationWarnings. 😄

PR #99, opened 2020-09-23
Add friendly names and docs to metrics definitions

tdsmith:
Follow up to https://github.com/mozilla/jetstream/issues/109.

@emtwo, this will be part of our solution for exposing documentation about metrics to the dashboarding layer eventually; is Markdown an okay format for the extended documentation strings?

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -185,10 +185,23 @@ class Metric:
     """Represents an experiment metric.
 
     Needs to be combined with an analysis window to be measurable!
+
+    Args:
+        name (str): A slug; uniquely identifies this metric in tables
+        data_source (DataSource): where to find the metric
+        select_expr (str): a SQL snippet representing a clause of a SELECT
+            expression describing how to compute the metric; must include an
+            aggregation function since it will be GROUPed BY client_id
+            and branch
+        friendly_name (str): A human-readable dashboard title for this metric
+        description (str): A paragraph of Markdown-formatted text describing
+            what the metric measures, to be shown on dashboards
     """
     name = attr.ib(type=str)
     data_source = attr.ib(type=DataSource)
     select_expr = attr.ib(type=str)
+    friendly_name = attr.ib(type=str, default="")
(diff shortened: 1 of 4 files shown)


tests/test_metric_libraries.py
```diff
@@ -9,29 +11,40 @@ def test_imported_ok():
     assert mmf.uri_count
 
 
-def test_sql_not_detectably_malformed():
-    for m in mmd.__dict__.values():
-        if isinstance(m, mmd.Metric):
-            sql_lint(m.select_expr.format(experiment_slug='slug'))
+def enumerate_included(klass):
```
felixlawrence, line 14: Nice!

src/mozanalysis/metrics/__init__.py
```diff
@@ -185,10 +185,23 @@ class Metric:
...
+            aggregation function since it will be GROUPed BY client_id
+            and branch
+        friendly_name (str): A human-readable dashboard title for this metric
+        description (str): A paragraph of Markdown-formatted text describing
+            what the metric measures, to be shown on dashboards
     """
     name = attr.ib(type=str)
     data_source = attr.ib(type=DataSource)
     select_expr = attr.ib(type=str)
+    friendly_name = attr.ib(type=str, default="")
```
felixlawrence, line 203: I'm a little uneasy about defaulting to "" rather than `None`. The latter better represents that no value was provided. I guess it's unlikely that someone will try to key a dictionary with `friendly_name` or count distinct `friendly_name` when `name` is available (in which case defaulting to `None` would save grief), but it's possible?

So I have a mild preference for changing this and description to default to `None`. If it's annoying to handle the `None` values downstream then I can live with the existing "".

felixlawrence (approved):
These files are in danger of becoming my go-to place for answering "how do I query that and what does it mean?" questions

src/mozanalysis/metrics/__init__.py
tdsmith: Oh, yeah, None is a fine choice here, idk why I chose `""`.

PR #100, opened 2020-09-24
Add friendly names and descriptions for segments

tdsmith:
Is there a good place to link people for more context on the segments? I was working from DTMO, which seemed kinda sparse.

### DIFF
--- a/src/mozanalysis/segments/__init__.py
+++ b/src/mozanalysis/segments/__init__.py
@@ -1,6 +1,8 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from typing import Optional
+
 import attr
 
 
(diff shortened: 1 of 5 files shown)


felixlawrence (approved):
Might be worth linking to each segment definition in each `description`? E.g. linking to https://docs.telemetry.mozilla.org/concepts/segments.html#regular-users-v3

Would love to hear specific pain points/requests for improvements for the docs.

tdsmith:
The DTMO copy describes them well but is there a good reference that talks about, like, how these were produced and why they're meaningful?

felixlawrence:
> The DTMO copy describes them well but is there a good reference that talks about, like, how these were produced and why they're meaningful?

Ah, most of that info is buried in the PRs and links in the PRs. E.g. [regular users v3](https://github.com/mozilla/firefox-data-docs/pull/466), [weekday/all-week regulars](https://github.com/mozilla/firefox-data-docs/pull/493). I thought "how these were produced" is mostly of historical interest. "Why they're meaningful" is going to be different in different contexts - e.g. for experiments, separating "new or resurrected v3" lets us (substantially) solve the problem where the length of the enrollment window determines the composition of the experiment population. "Write about how they can be applied to all the different contexts" seemed like too-open-ended a task when there was no guarantee about which contexts there was actually going to be work done, in particular work done that could benefit from use of the segments...

Happy to discuss further over slack (or here, if you prefer)

tdsmith:
Okay, thanks. If we develop any new guidance for experimenters we can add it here :)

felixlawrence:
The guidance is _cough_ baked into mozanalysis: you determine people's segments exclusively based on data collected before they enroll in the experiment, to avoid bias from the experiment's treatment making users more/less likely to be measured as being in a given segment.

If we add country-based segments to mozanalysis (probably should!) we might break this rule slightly? Unless we can get the country out of the ping that gave us the enrollment event.

tdsmith:
Sure, but with respect to when and how they should be interested in an analysis of their metrics with respect to any specific segment.

> we might break this rule slightly?

Yeah, I faked something similar here: https://github.com/mozilla/jetstream-config/blob/main/bug-1647305-pref-comcast-steering-experiment-release-78-80.toml#L67-L83

felixlawrence:
Yeah; it may even be worth adjusting our enrollment queries to make this information available as a column in `raw_enrollments` and support a None DataSource...

PR #101, opened 2020-10-06
Emit Sphinx docs for desktop metrics

tdsmith:
It turns out Sphinx refuses to emit anything for undocumented module constants. But we can add trivial documentation.

If anyone gets bored enough, I think the proper thing to do here is to write a Sphinx extension that hooks the autodoc [process-docstring](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#docstring-preprocessing) event and replaces "..." with the `description` field, though I'm not sure how to reason about the description containing Markdown and the Sphinx docs being more or less in RST -- eh!

### DIFF
--- a/docs/api/metrics/desktop.rst
+++ b/docs/api/metrics/desktop.rst
@@ -1,4 +1,5 @@
 :mod:`mozanalysis.metrics.desktop`
 ----------------------------------
 
-See the source for the included metrics. Sphinx seems to be uncooperative displaying source code that does not define functions or classes - so view the latest source `on github <https://github.com/mozilla/mozanalysis/tree/master/src/mozanalysis/metrics/desktop.py>`_.
+.. automodule:: mozanalysis.metrics.desktop
+    :members:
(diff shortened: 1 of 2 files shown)


felixlawrence (approved):
For the `Metric`s, we already have a `description`, which is there as documentation. Is there some way to re-use this as the documentation for sphinx?

I was hoping that we could stick the description into a docstring, and then change our calls to `metric.description` to `metric.__doc__`, but [that's not possible](https://stackoverflow.com/questions/8820276/docstring-for-variable)...

The current proposal is a step forward and improves the documentation, but every person reading the documentation for metrics is going to wonder why it looks weird, and fixing this weirdness won't be ROI+. My inner pedant cries.

tdsmith:
> Is there some way to re-use this as the documentation for sphinx?

Yeah, that was where I was going with my rambling about Sphinx plugins. It seems possible and doesn't look like a lot of code but I'm sure it'll take a couple hours to figure out how to hook it up right. Someday :p

PR #102, opened 2020-10-20
Add metrics for pocket tile clicks

felixlawrence:
@kirill-demtchouk has applied these to an experiment; I wrote them but haven't personally vetted them.


kirill-demtchouk:
Looks good to me! Thanks, @felixlawrence !

tdsmith (approved):
lgtm, though I'm realizing that my choice of "Counts [...]" for binary metrics was an odd choice. We can fix them all later though!

PR #104, opened 2020-11-12
Add a new-profile segment

tdsmith:
Use clients_first_seen to identify profiles that enroll on their first date of use.

### DIFF
--- a/src/mozanalysis/segments/desktop.py
+++ b/src/mozanalysis/segments/desktop.py
@@ -15,6 +15,14 @@ clients_last_seen = SegmentDataSource(
 )
 
 
+clients_first_seen_today = SegmentDataSource(
+    name="clients_first_seen_today",
+    from_expr="`moz-fx-data-shared-prod`.telemetry.clients_first_seen",
+    window_start=0,
+    window_end=0,
+    submission_date_column="first_seen_date",
+)
+
 regular_users_v3 = Segment(
     name='regular_users_v3',
     data_source=clients_last_seen,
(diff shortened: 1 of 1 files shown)


felixlawrence (approved):
LGTM.

There are going to be errors around `clients_first_seen` dealing with the `main` ping while the enrollment event is in the `events` ping, so this might exclude a fraction of new users who enroll on day 1.

It might be worth checking this segment against e.g. an onboarding experiment, to see what fraction of new users we're missing. If it's too high, we might want to mitigate the situation by setting `window_start=-1`. `mozanalysis` won't let us set `window_end=1` because it could introduce bias; if the data shows a strong need to change this for this segment then we should do so.

I think that the segment as expressed in this PR is likely to be useful regardless of the outcome of the above checks that I recommend, so I don't think they should necessarily block merging this.

tdsmith:
Thanks for the comments. I think that's worth documenting -- I'll hold off a little on merging that until I pull those numbers.

jklukas:
Note that `first_seen_date` is now planned to be available directly in `clients_last_seen`. Should land the first week of December: https://github.com/mozilla/bigquery-etl/pull/1561

tdsmith:
Felix pointed out that we often receive events from users before we receive a main ping from them, and those differences can straddle UTC midnight, which is confirmed at https://sql.telemetry.mozilla.org/queries/76784/source#191390 -- but if we haven't received a main ping from them by the time they enroll, it's probably reasonable to conclude that they're new.

Hence the allowances for the impossible—first_seen_date cannot be greater than submission_date in that table—mostly for emphasis.

PR #105, opened 2020-11-18
Add a days of use metric for desktop

tdsmith:


felixlawrence (approved):

tdsmith:
@felixlawrence two questions for you actually -- 
- is a ping-based definition consistent with the KPI you think we'll end up with, and
- any idea what's up with this test failure? I'll dig but if it's obvious to you then I won't :D

src/mozanalysis/metrics/desktop.py
```diff
@@ -299,3 +299,13 @@
         Counts the number of pocket tile clicks made by each client.
     """),
 )
+
+
+#: Metric: ...
+days_of_use = Metric(
+    name="days_of_use",
+    data_source="clients_daily",
```
tdsmith, line 307: oh, lol

felixlawrence:
> is a ping-based definition consistent with the KPI you think we'll end up with, and

Yes, I think so? I think there is a push for "core actives" to only include >1 uri users, but [this is a bad idea IMO](https://docs.google.com/document/d/1v2awThGc8rBLDd6j93cvFvy_i1dRPuzqZq57rrktQK8/edit#). I'd be amenable to ">0 active hours", but the simplest thing to do is just watch "sent a ping" and [monitor the other usage criteria](https://colab.research.google.com/drive/1K3n6KlXORMLlScTyFphI2l01Z7K-PbB1#scrollTo=OGjZuXYqTDzK) so we know we're not optimising for users that don't use the product.

felixlawrence:
I.e. DAU is still calculated from "sent a main ping"

tdsmith:
Awesome, thanks!

PR #106, opened 2020-11-18
Add a lower-is-better flag

tdsmith:
Dashboards shouldn't look excited when we get more of a bad thing, or when performance timings increase.

Sometimes this will depend on the context of the experiment; if the wrong assertion is made here, folks can make a copy of the metric with the bit flipped.

cf https://github.com/mozilla/jetstream/issues/252


felixlawrence (approved):
r+ to the idea, but can we avoid what's almost a double-negative and flip this around, defaulting to `True`? E.g. `positive_is_positive` (ew) or `higher_is_better`?

(i've fallen into this very trap before :D )

PR #107, opened 2020-11-19
Fix the `days_of_use` metric.

felixlawrence:
@SuYoungHong noticed that the results of the `days_of_use` metric introduced in PR #105 were very suspicious.

The `COUNT(*)` was counting enrollments even when they had no row in `clients_daily`. Oops.

Instead, we need to `COUNT()` a non-null column from `clients_daily`; `COUNT(ds.submission_date)` seems to be the best for communicating what this metric does? And it's not worth adding a `DISTINCT` because it's implied by `clients_daily` and the lack of other joins?

I haven't tried this code yet but I've asked Su to do so in [this notebook](https://colab.research.google.com/drive/1E-f35o3DE6FIdnmNjs7eEA413lkNjtLh).

### DIFF
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -306,7 +306,7 @@ pocket_click_count = Metric(
 days_of_use = Metric(
     name="days_of_use",
     data_source=clients_daily,
-    select_expr="COUNT(*)",
+    select_expr="COUNT(ds.submission_date)",
     friendly_name="Days of use",
     description="The number of days in the interval that each client sent a main ping."
 )


tdsmith:
lol, right: joins.

i think some of the fenix metrics make this mistake as well. `pings_aggregated_by_this_row` is a clients_daily-only column that avoids having secret knowledge about the construction of the query, maybe let's prefer that

felixlawrence:
It's an ugly tradeoff - count some bizarre row that you need secret knowledge about the ETL to know that the column is not null and that's why it's being chosen, or know how mozanalysis constructs its queries. I significantly prefer the latter because the truth is in the same codebase rather than "out there"; it should be less brittle?

If this is disagreeable, another option that would remove this potential trap would be to turn the join in `DataSource.build_query()` into an `INNER JOIN` and rely on `Experiment._build_metrics_query_bits()` using a `LEFT JOIN`, which it does. I didn't immediately run to this option because there's higher potential for it to break things.

felixlawrence:
Yeah maybe we should switch to the `INNER JOIN` in `DataSource.build_query()`; given the erroneous metrics for Fenix, this appears to be a footgun, which we can disarm.

I'm not going to make this change today, because we should check it carefully (i.e. run experiment analyses with lots of metrics and both versions of the join, and check that the numbers are consistent). I might have time to do it on the Canadian Productivity Days late next week, otherwise it'd be awesome if you could have a go.

tdsmith (approved):
> It's an ugly tradeoff - count some bizarre row that you need secret knowledge about the ETL to know that the column is not null and that's why it's being chosen, or know how mozanalysis constructs its queries. I significantly prefer the latter because the truth is in the same codebase rather than "out there"; it should be less brittle?

I guess it doesn't really matter -- we don't expect clients to look at these definitions as long as they do what they say they do (oops) -- but this punches through an abstraction and asserts a new contract in a way that depending on a column in a table doesn't. We're always responsible for understanding the tables we're operating on.

But this seems fine to merge.

felixlawrence:
Given that there are two independent(?) instances of this error in the codebase, I think we should make the `INNER JOIN` switch. That is now #108. In the meantime, let's merge this fix for this metric.

PR #109, opened 2020-11-20
Materialize enrollments table

tdsmith:
Repeated references to the enrollments CTE seem to stress BigQuery out, giving rise to "Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex." errors.

If we use CREATE TEMPORARY TABLE to materialize the enrollments table before joining it everywhere, we get a scripting query with similar semantics and a calmer query planner.

~Caveat: I just realize this doesn't quite work yet, so leaving this as a draft for now; we can't set a destination-table name at execution time for a scripting query, so we have to replace the SELECT down there with another `CREATE TABLE {destination_table} AS` instead.~

cc @felixlawrence, @scholtzan

### DIFF
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import re
 from google.cloud import bigquery
-from google.api_core.exceptions import Conflict, NotFound
+from google.api_core.exceptions import NotFound
 
 
 def sanitize_table_name_for_bq(table_name):
@@ -31,33 +31,6 @@ class BigQueryContext:
         self.project_id = project_id
         self.client = bigquery.Client(project=project_id)
 
-    def run_query(self, sql, results_table=None):
-        """Run a query and return the result.
-
-        If ``results_table`` is provided, then save the results
-        into there (or just query from there if it already exists).
-
-        Returns a ``google.cloud.bigquery.table.RowIterator``
-        """
-        if not results_table:
-            return self.client.query(sql).result()
-
-        try:
-            full_res = self.client.query(
-                sql,
-                job_config=bigquery.QueryJobConfig(
-                    destination=self.client.dataset(
-                        self.dataset_id
-                    ).table(results_table)
-                )
-            ).result()
-            print('Saved into', results_table)
-            return full_res
-
-        except Conflict:
-            print("Full results table already exists. Reusing", results_table)
-            return self.client.list_rows(self.fully_qualify_table_name(results_table))
-
     def run_script_or_fetch(self, sql, results_table):
         """Runs a BigQuery SQL script and returns a RowIterator for results_table.
         The script is assumed to create a table named results_table after completing
@@ -74,11 +47,14 @@ class BigQueryContext:
         fqtn = self.fully_qualify_table_name(results_table)
 
         try:
-            return self.client.list_rows(fqtn)
+            cached = self.client.list_rows(fqtn)
+            print("Full results table already exists. Reusing", results_table)
+            return cached
         except NotFound:
             pass
 
         self.client.query(sql).result()
+        print('Saved into', results_table)
         return self.client.list_rows(fqtn)
 
     def fully_qualify_table_name(self, table_name):
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import re
 from google.cloud import bigquery
-from google.api_core.exceptions import Conflict
+from google.api_core.exceptions import Conflict, NotFound
 
 
 def sanitize_table_name_for_bq(table_name):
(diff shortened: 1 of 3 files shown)


scholtzan:
This looks good to me 👍

src/mozanalysis/experiment.py
```diff
@@ -173,16 +173,29 @@ def get_single_window_data(
             analysis_length_days, self.num_dates_enrollment
         )
 
-        sql = self.build_query(
-            metric_list, time_limits, enrollments_query_type, custom_enrollments_query,
-            segment_list
+        sql_to_be_hashed = self.build_query(
+            metric_list=metric_list,
+            time_limits=time_limits,
+            destination_table="",
```
felixlawrence, line 179: I'm kind of disappointed that you didn't go for a crypto-mining like solution, where you need to keep trying different values here until you find one that matches the hash.

src/mozanalysis/experiment.py
```diff
@@ -313,29 +341,35 @@ def build_query(
...
+        enrollments AS (
+            SELECT
+                e.*,
+                aw.*
+            FROM segmented_enrollments e
+            CROSS JOIN analysis_windows aw
+        )
+    );
+
+    CREATE OR REPLACE TABLE {destination_table} AS (
```
felixlawrence, line 359: Is `CREATE OR REPLACE` a good idea? If this table is being replaced rather than created, is that a sign that something is going wrong (i.e. we're not caching properly)?

Or would that be "too clever" and bite us in the bum when we're trying to re-run mozanalysis after a failed ETL run?

src/mozanalysis/experiment.py
```diff
@@ -313,29 +341,35 @@ def build_query(
         )
 
         return """
-    WITH analysis_windows AS (
-        {analysis_windows_query}
-    ),
-    raw_enrollments AS ({enrollments_query}),
-    segmented_enrollments AS ({segments_query}),
-    enrollments AS (
+    CREATE TEMPORARY TABLE enrollments AS (
```
felixlawrence, line 344: Can you remind me how temporary `TEMPORARY` is? Might multiple simultaneous instances of mozanalysis clobber each other? Might sequentially running jobs hit issues?

felixlawrence (commented):
This breaks backward compat (not a biggie, but not preferred), and makes things more complicated:
* "Should I use `run_query()` or `run_script_or_fetch()`? What's the difference?"
* Generally, power users need to remix `Experiment.get_single_window_data()` and `Experiment.get_time_series_data()` (replacing one or more of the calls - e.g. custom `time_limits`), so it's good to keep them as short and simple as possible

I think, with one white lie, it might be possible to add your functionality more elegantly. If the variable formerly known as `sql` contains a string that's _almost_ your SQL script - the just has the following line in it - then I think we can do things more simply:

```     CREATE OR REPLACE TABLE {destination_table} AS (```

(i.e., you would revert `build_query()`'s function signature to the existing version, and add `destination_table='{destination_table}'` to the `format()` call.

Then in `BigQueryContext.run_script_or_fetch()`, we would call 
`self.client.query(sql.format(destination_table=fqtn)).result()`... and possibly be able to replace `BigQueryContext.run_query()` with this new version as a drop-in replacement?

Might be good not to introduce the inconsistency of `results_table` vs `destination_table`. Unless you really hate `results_table`, I suggest it should win since it's there already; if you do hate it then go change everything over to `destination_table`.

File: src/mozanalysis/experiment.py (hunk 3 above)
Comment (tdsmith) [reply src/mozanalysis/experiment.py:344]: Extremely; they're not accessible except from the script that creates them. See https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#temporary_tables

File: src/mozanalysis/experiment.py (hunk 2 above)
Comment (tdsmith) [reply src/mozanalysis/experiment.py:359]: This is useful for Jetstream because we re-use names and don't want the caching behavior.

**Comment by tdsmith** [comment]
>   * "Should I use `run_query()` or `run_script_or_fetch()`? What's the difference?"

I left a note about this in the commit message but the reason this is an important distinction is that it's illegal to specify a destination table when executing a script, so we do need to execute them in different ways, so I think we shouldn't merge these.

> I think, with one white lie, it might be possible to add your functionality more elegantly. If the variable formerly known as `sql` contains a string that's _almost_ your SQL script - the just has the following line in it - then I think we can do things more simply:
> 
> ` CREATE OR REPLACE TABLE {destination_table} AS (`
> 
> (i.e., you would revert `build_query()`'s function signature to the existing version, and add `destination_table='{destination_table}'` to the `format()` call.

This changes the build_query contract to return a Python template string instead of a query—I don't love string templates as a return type but I don't have a good argument about it. Changing the name of the function to build_query_template would at least make the incompatible change obvious.

I guess I'm surprised that people are rewriting Experiment.get_*_window_data().

> Might be good not to introduce the inconsistency of `results_table` vs `destination_table`.

You're right; I should match what's there. I had the BigQuery API term stuck in my head.

File: src/mozanalysis/bq.py [hunk 4]
```diff
@@ -62,6 +62,29 @@ def run_query(self, sql, results_table=None):
...
+        destination_table is assumed to be an unqualified table name without
+        a project or dataset reference.
+
+        Learn more about BigQuery scripting at
+        https://cloud.google.com/bigquery/docs/reference/standard-sql/scripting.
+        """
+
+        fqtn = self.fully_qualify_table_name(destination_table)
+
+        try:
```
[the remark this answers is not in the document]
tdsmith: Sure! I was just going for the minimum change.

felixlawrence (approved):
Great! Thanks.

PR #110, opened 2020-11-28
Build hacks

tdsmith:
- Declare pyarrow and google-cloud-bigquery-storage as dependencies so people get fast .to_dataframe() without extra work
- Build on interesting Python environments: Colab is 3.6 and Jetstream uses 3.8, so test those


felixlawrence (approved):
Oooft, the upgrade treadmill.

PR #111, opened 2020-12-01
Revise metrics for Pocket tile clicks.

kirill-demtchouk:
Revise the Pocket tile clicks metrics added in an earlier PR:
https://github.com/mozilla/mozanalysis/pull/102

In particular, make two changes:
-drop the metric for whether a user clicked a Pocket tile at all in favor of only count metrics
-split "Pocket tile clicks" into two buckets: Pocket organic recommendation clicks and Pocket sponsored content clicks

### DIFF
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -272,35 +272,35 @@ connect_fxa = Metric(
     """),
 )
 
-#: Metric: ...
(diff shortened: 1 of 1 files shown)


kirill-demtchouk:
@felixlawrence @tdsmith Could either of you give this a review when you have a chance? Thank you!

tdsmith (approved):

Issue #108, opened 2020-11-20
Use INNER JOIN in DataSource.build_query()


felixlawrence:
In #107 we encountered the second instance of a footgun: we used a `COUNT(*)` in a Metric, and this was bad because we're left-joining from the list of enrollments so the `COUNT(*)` evaluates to 1 for rows where no row existed in the `DataSource`.

> another option that would remove this potential trap would be to turn the join in DataSource.build_query() into an INNER JOIN and rely on Experiment._build_metrics_query_bits() using a LEFT JOIN, which it does. I didn't immediately run to this option because there's higher potential for it to break things.

We should check whether this is a good idea, and if so, do it.

PR #113, opened 2020-12-07
Fix ping count metrics

tdsmith:
Addresses the issue described in #107, where the join changes the meaning of COUNT(*).


felixlawrence (approved):
I note that this uses the opposite strategy to #107 (find a field we know is unique to the table); would have preferred to be consistent. But ultimately I don't mind because I see these as being relatively short-lived until we do #108?

`document_id` is a good safe choice that might drop enough of a hint to the astute reader that they can work out what's happening?

tdsmith:
Thanks!

PR #114, opened 2020-12-07
Make metric datasets configurable at runtime

tdsmith:
Different release channels of Glean apps have their data stored in different BigQuery datasets. This departs from the legacy telemetry practice on desktop, where data from all release channels end up in a single table.
This means that, although it's reasonable to talk about a set of metrics that are valid across the different Fenix release channels, the underlying DataSources need to be customized for each release channel.
I had once expected that this would be easy to do just by defining an alternate set of DataSources and rebasing the Metrics onto the new DataSources -- but the `events` data source is a good example of why this isn't as easy as I'd like. Data sources can be complex and it would be annoying to redefine them.

Instead, allow DataSources to be templated over dataset, and have the Experiment class forward some information from the client about which dataset the metrics should be based on.

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -43,7 +43,7 @@ class DataSource:
             `{dataset}` parameter.
     """
     name = attr.ib(validator=attr.validators.instance_of(str))
-    from_expr = attr.ib(validator=attr.validators.instance_of(str))
+    _from_expr = attr.ib(validator=attr.validators.instance_of(str))
     experiments_column_type = attr.ib(default='simple', type=str)
     client_id_column = attr.ib(default='client_id', type=str)
     submission_date_column = attr.ib(default='submission_date', type=str)
@@ -64,15 +64,22 @@ class DataSource:
         self.from_expr_for(None)
 
     def from_expr_for(self, dataset: Optional[str]) -> str:
+        """Expands the ``from_expr`` template for the given dataset.
+        If ``from_expr`` is not a template, returns ``from_expr``.
+
+        Args:
+            dataset (str or None): Dataset name to substitute
+                into the from expression.
+        """
         effective_dataset = dataset or self.default_dataset
         if effective_dataset is None:
             try:
-                return self.from_expr.format()
+                return self._from_expr.format()
             except Exception as e:
                 raise ValueError(
                     f"{self.name}: from_expr contains a dataset template but no value was provided."  # noqa:E501
                 ) from e
-        return self.from_expr.format(dataset=effective_dataset)
+        return self._from_expr.format(dataset=effective_dataset)
 
     @property
     def experiments_column_expr(self):
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -14,7 +14,10 @@ class DataSource:
         name (str): Name for the Data Source. Used in sanity metric
             column names.
         from_expr (str): FROM expression - often just a fully-qualified
-            table name. Sometimes a subquery.
+            table name. Sometimes a subquery. May contain the string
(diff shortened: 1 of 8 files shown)


tdsmith:
fyi @scholtzan

src/mozanalysis/metrics/__init__.py
```diff
@@ -51,6 +59,28 @@ def _check_experiments_column_type(self, attribute, value):
...
+
+    def from_expr_for(self, dataset: Optional[str]) -> str:
+        """Expands the ``from_expr`` template for the given dataset.
+        If ``from_expr`` is not a template, returns ``from_expr``.
+
+        Args:
+            dataset (str or None): Dataset name to substitute
+                into the from expression.
+        """
+        effective_dataset = dataset or self.default_dataset
```
felixlawrence, line 74: Do we _need_ to be this friendly with the error messages? If not, `.format()` has got behaviour pretty close to what we need?
```
kwargs = {'dataset': effective_dataset} if effective_dataset else {}
return self._from_expr.format(**kwargs)
```

src/mozanalysis/metrics/__init__.py
```diff
@@ -83,7 +113,13 @@ def experiments_column_expr(self):
         else:
             raise ValueError
 
-    def build_query(self, metric_list, time_limits, experiment_slug):
+    def build_query(
+            self,
```
felixlawrence, line 117: I think our indentation style is to dedent the list of arguments once, and to dedent the `):` twice

src/mozanalysis/metrics/__init__.py
```diff
@@ -206,14 +242,6 @@ class Metric:
     description = attr.ib(type=Optional[str], default=None)
     bigger_is_better = attr.ib(type=bool, default=True)
 
-    def from_data_source(self, data_source: DataSource) -> "Metric":
```
felixlawrence, line 209: I think I agree with deleting this. On the one hand we may want to use this for glean apps on different platforms. On the other hand, copypasta in this context is probably safer because a Fenix data scientist could unknowingly break the iOS metrics.

src/mozanalysis/segments/__init__.py
```diff
@@ -33,15 +36,48 @@ class SegmentDataSource:
...
+            try:
+                return self._from_expr.format()
+            except Exception as e:
+                raise ValueError(
+                    f"{self.name}: from_expr contains a dataset template but no value was provided."  # noqa:E501
+                ) from e
+        return self._from_expr.format(dataset=effective_dataset)
+
+    def build_query(
+            self,
```
felixlawrence, line 75: Indentation here too

tests/test_experiment.py
```diff
@@ -289,6 +292,44 @@ def test_segments_megaquery_not_detectably_malformed():
...
+
+    tl = TimeLimits.for_ts(
+        first_enrollment_date='2019-01-01',
+        last_date_full_data='2019-03-01',
+        time_series_period='weekly',
+        num_dates_enrollment=8
+    )
+
+    sds = SegmentDataSource(
+        name="cool_data_source",
```
felixlawrence, line 306: :sunglasses:

felixlawrence: Why does no-one ever pick `daggy_data_source`

felixlawrence (approved):
Sensible direction! Glad for the validator.

src/mozanalysis/metrics/__init__.py
tdsmith: Yeah I thiiiiiiiiink it just works out to be an attractive nuisance but it's easy enough to revert if we want it back.

src/mozanalysis/metrics/__init__.py
tdsmith: I was tempted but I don't think we'll regret the verbosity even if it adds a few loc.

PR #115, opened 2020-12-08
Blacken

tdsmith:
Downsides:
* Big annoying diff
* Some decisions look a little silly
* A boot stamping on a human face - forever

Upsides:
* Set your editor to auto-blacken on save and never think about formatting again

It's made hacking on jetstream more pleasant; would you mind if we adopted it here?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,33 +7,38 @@
 from setuptools import setup, find_packages
 
 tests_require = [
-    'mock', 'pytest', 'pytest-cov', 'pytest-timeout', 'pyspark'
+    "mock",
+    "pytest",
+    "pytest-black",
+    "pytest-cov",
+    "pytest-timeout",
+    "pyspark",
 ]
 
-docs_require = ['Sphinx', 'sphinx-autobuild', 'sphinx-rtd-theme']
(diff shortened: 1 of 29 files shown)


felixlawrence (approved):
The occasional silly decision is a small price to pay to be free from having to make decisions (oof, do I now love big brother?).

I'm not thrilled about a formatting tool that's incompatible with our linter. Is there some config that lets us change the linter or the formatter? Or can we break up the long strings to avoid the dreaded `# noqa`?

Is there some way we can add Black as a pre-commit hook or equivalent, so that we don't rely on all contributors manually configuring their text editors? Or without a tool like Phabricator's Arcanist, would our best option be having a CI job to run Black and check it's a no-op - which is overkill for us given that it's not thaat helpful.

tdsmith:
The noqa:E501's predate black; I just had to shuffle them around after the reformatting.

I think the noqa's are fine; black will make the right decision about line lengths for code, and E501 remains as a reminder not to let docstrings get too wide, and we can shut it off when it would make us jump through dumb hoops.

There's no way to make pre-commit hooks run in github; folks would still have to set that up locally for the project. We can use pytest-black to make CI fail if black would change something, though, which is probably worth doing here.

PR #116, opened 2020-12-10
Rewrite metrics and enrollments query

scholtzan:
This is an exploration for fixing https://github.com/mozilla/jetstream/issues/357

I shouldn't have run black on the files I changed...

I haven't tested this yet, just wanted to check if this approach is acceptable.

### DIFF
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -79,9 +79,7 @@ class BigQueryContext:
 
         except Conflict:
             print("Full results table already exists. Reusing", results_table)
-            return self.client.query(
-                "SELECT * FROM {}".format(self.fully_qualify_table_name(results_table))
-            ).result()
+            return self.client.list_rows(results_table)
 
     def fully_qualify_table_name(self, table_name):
         """Given a table name, return it fully qualified."""
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -36,26 +36,6 @@ class BigQueryContext:
         self.project_id = project_id
         self.client = bigquery.Client(project=project_id)
 
-    def run_query(self, sql, results_table=None) -> str:
-        """Runs a SQL query and returns the destination table.
-        If ``results_table`` is provided, then save the results
-        into there otherwise save results in a temporary table.
-        """
-        if results_table:
-            job = self.client.query(
-                sql,
-                job_config=bigquery.QueryJobConfig(
-                    destination=self.client.dataset(self.dataset_id).table(
-                        results_table
-                    )
-                ),
-            )
-        else:
-            job = self.client.query(sql)
-
-        job.result()
-        return job.destination
-
     def run_query_and_fetch(self, sql, results_table=None):
         """Run a query and return the result.
         If ``results_table`` is provided, then save the results
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import re
 
-from google.api_core.exceptions import NotFound
+from google.api_core.exceptions import Conflict
 from google.cloud import bigquery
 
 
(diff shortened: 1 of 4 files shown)


src/mozanalysis/experiment.py
```diff
@@ -250,57 +265,63 @@ def get_time_series_data(
...
+                [
+                    last_date_full_data,
+                    self.experiment_slug,
+                    hash_ish(sql_template_metrics),
+                ]
+            )
+        )
 
-        bq_context.run_script_or_fetch(sql, full_res_table_name)
+        bq_context.run_query_and_fetch(sql_template_metrics, full_res_table_name)
```
scholtzan, line 299: This is more or less what jetstream will do. The enrollments and metrics query can then be validated in a dry run

src/mozanalysis/experiment.py
```diff
@@ -322,45 +343,69 @@ def build_query_template(
...
+            segments_query=segments_query,
+        )
 
-        SELECT
-            e.*,
-            aw.*
-        FROM segmented_enrollments e
-        CROSS JOIN analysis_windows aw
-    );
+    def build_metrics_query(
```
scholtzan, line 370: Separate function for generating the metrics query.

src/mozanalysis/experiment.py
```diff
@@ -250,57 +265,63 @@ def get_time_series_data(
...
         return TimeSeriesResult(
             fully_qualified_table_name=bq_context.fully_qualify_table_name(
                 full_res_table_name
             ),
-            analysis_windows=time_limits.analysis_windows
+            analysis_windows=time_limits.analysis_windows,
         )
 
-    def build_query_template(
+    def build_enrollment_query(
```
scholtzan, line 308: Separate function for generating the enrollments query.

tdsmith:
Checking my understanding:
- The problem is that BigQuery isn't really performing validation because these queries are currently executed as a script instead of simple queries and BigQuery doesn't execute scripts while dry-running (oops)
- You're proposing to solve this by materializing enrolments into a concrete table in a separate query from the metrics query, and executing them separately, so that jetstream can dry-run them separately

I think that sounds reasonable.

One fun fact for Jetstream is that the enrollments query should never change after an experiment's enrollment period ends (unless segments are added by configuration changes), so it's perfectly reasonable to materialize this somewhere once and then rely on it forever. We can take advantage of this for interactive use in mozanalysis as well by following the existing content-addressable table name pattern.

That'll make queries somewhat faster and save a few dollars on data access, and being able to inspect the enrolments table directly is nice for diagnosis.

@felixlawrence, you had some concerns about increasing the complexity of these Experiment functions. The lack of validation for the script query is really a non-starter for the jetstream-config UX (direct quote from yours truly: "Jesus landing a config is such a pain in the ass"), so we're very motivated to do something here. 😄 Does this seem like a good tradeoff?

I can think of some alternatives where interactive use and Jetstream use different code paths here but I think splitting this query up probably serves everyone's interests.

src/mozanalysis/bq.py
```diff
@@ -31,31 +32,42 @@ def __init__(self, dataset_id, project_id='moz-fx-data-bq-data-science'):
...
+            ).result()
+            print('Saved into', results_table)
+            return full_res
+
+        except Conflict:
             print("Full results table already exists. Reusing", results_table)
-            return cached
-        except NotFound:
-            pass
+            return self.client.query(
```
tdsmith, line 65: `self.client.list_rows(table name)` is identical and a little more concise

src/mozanalysis/experiment.py
```diff
@@ -254,21 +254,26 @@ def get_time_series_data(
...
+        sql_template_enrollments = self.build_enrollment_query(
             time_limits=time_limits,
             enrollments_query_type=enrollments_query_type,
             custom_enrollments_query=custom_enrollments_query,
             segment_list=segment_list,
         )
 
+        enrollments_table = bq_context.run(sql_template_enrollments)
+
+        sql_template_metrics = self.build_metrics_query(
```
tdsmith, line 266: This is no longer a template, right?

scholtzan:
Just to clarify, running the enrollments query will write the results in a temporary table (in a special dataset with some random identifier) which will get deleted at some point.

src/mozanalysis/experiment.py
scholtzan: Yeah, the naming here needs some adjustment

tdsmith:
Yeah, I see that this PR lets BQ choose a temporary name. I think explicitly persisting the enrollments tables would be a valuable enhancement for all consumers :)

felixlawrence:
Yeah, my preference to keep the user-callable functions simple is strongly outweighed here. Go for it.

src/mozanalysis/experiment.py
```diff
@@ -176,21 +176,29 @@ def get_single_window_data(
...
+            [last_date_full_data, self.experiment_slug, hash_ish(metrics_sql)]
         ))
 
-        sql = sql_template.format(results_table=full_res_table_name)
-
-        return bq_context.run_script_or_fetch(sql, full_res_table_name).to_dataframe()
+        return bq_context.run_query_and_fetch(
+            metrics_sql,
+            full_res_table_name
+        ).to_dataframe()
```
scholtzan, line 201: Fixes https://github.com/mozilla/mozanalysis/issues/119

scholtzan:
All blackened and ready for review

src/mozanalysis/experiment.py
```diff
@@ -271,23 +276,26 @@ def get_time_series_data(
...
             enrollments_query_type=enrollments_query_type,
             custom_enrollments_query=custom_enrollments_query,
             segment_list=segment_list,
         )
 
-        full_res_table_name = sanitize_table_name_for_bq(
-            "_".join(
-                [last_date_full_data, self.experiment_slug, hash_ish(sql_template)]
-            )
+        enrollments_table = bq_context.run_query(enrollments_sql)
```
tdsmith, line 286: Let's name this table! We can use the same

```
        full_res_table_name = sanitize_table_name_for_bq(
            "_".join([last_date_full_data, self.experiment_slug, hash_ish(metrics_sql)])
        )
```

strategy we use below.

tdsmith: Then we can use run_query_and_fetch everywhere.

tdsmith (approved):
Looks good!

I'd like to name and persist the enrollments table, and maybe refactor the core query logic in get_whatever_data into a private function, but I could take that as followup work.

PR #120, opened 2020-12-24
Join enrollments more rigorously... and fix what was broken

### LINKED ISSUE #117: build_query doesn't respect enrollments table contract
When calculating Days of Use metric for a given window period, there are instances of clients producing more Days of Use then possible given the window period. 

If window is 14 days, only 14 days of use is possible. However, we see cases of greater then 14 days. 

Interesting observation: the maximum number of days of use we see seem to always be: 
* (# of branches) x (possible days)

Example: 

pulled data using: [this notebook](https://colab.research.google.com/drive/1LY_cCOXepUHxt5oAUKDvFGtG2RB1RL4d#scrollTo=_TeZxV3r2k3d&line=10&uniqifier=1)

```
use_days_wk0wk1 = exp.get_single_window_data(
    bq_context=bq_context,
    metric_list=[
        mmd.days_of_use, 
        imported_items, 
        imported_logins, 
        enabled_themes
    ],
    last_date_full_data='2020-11-18',
    analysis_start_days=0,
    analysis_length_days=14
)
```

Table here: `moz-fx-data-bq-data-science.shong.2020_11_18_bug_1668928_message_multi_stage_aboutwelcome_zero_stage_release_82_83_0c385114523c`

Example of issue: 

```
select *
from `moz-fx-data-bq-data-science.shong.2020_11_18_bug_1668928_message_multi_stage_aboutwelcome_zero_stage_release_82_83_0c385114523c`
WHERE days_of_use > 14
```

and 

```
select days_of_use, count(*)
from `moz-fx-data-bq-data-science.shong.2020_11_18_bug_1668928_message_multi_stage_aboutwelcome_zero_stage_release_82_83_0c385114523c`
GROUP BY 1
```

felixlawrence:
When I went to test in [a notebook](https://colab.research.google.com/drive/1x9wBQc7h80zpmymH-gs7seoP2Opxd-QG) the original change that fixes #117 and an equivalent bug with segments (commit 56d97bd), I found a bunch of apparently-broken functionality that led me on a merry end-of-year yak shave. Thanks @tdsmith for coming along for the ride!

In [the notebook](https://colab.research.google.com/drive/1x9wBQc7h80zpmymH-gs7seoP2Opxd-QG) I tested both `Experiment.get_time_series_data()` and `Experiment.get_single_window_data()`, and the results indicate that the bug is fixed (max 7 days of use in a week) and the methods both work and yield consistent results.

We should consider making the result _formats_ more consistent (hiding client_id and analysis window details from get_single_window_data), but that's an issue for another day.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -191,25 +191,27 @@ class Experiment:
             segment_list=segment_list,
         )
 
-        enrollments_table_name = sanitize_table_name_for_bq(
-            "_".join(
-                [
-                    last_date_full_data,
-                    "enrollments",
-                    self.experiment_slug,
-                    hash_ish(enrollments_sql),
-                ]
+        enrollments_table_name = bq_context.fully_qualify_table_name(
+            sanitize_table_name_for_bq(
+                "_".join(
+                    [
+                        last_date_full_data,
+                        "enrollments",
+                        self.experiment_slug,
+                        hash_ish(enrollments_sql),
+                    ]
+                )
             )
         )
 
-        enrollments_table = bq_context.run_query_and_fetch(
+        bq_context.run_query_and_fetch(
             enrollments_sql, enrollments_table_name
         )
 
         metrics_sql = self.build_metrics_query(
             metric_list=metric_list,
             time_limits=time_limits,
-            enrollments_table=enrollments_table,
+            enrollments_table=enrollments_table_name,
         )
 
         full_res_table_name = sanitize_table_name_for_bq(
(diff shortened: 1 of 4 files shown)


src/mozanalysis/experiment.py
```diff
@@ -366,12 +376,14 @@ def build_enrollments_query(
                 {analysis_windows_query}
             ),
             raw_enrollments AS ({enrollments_query}),
+            -- TODO: enforce that raw_enrollments is uniquely keyed
+            -- by (client_id, branch)
```
tdsmith, line 380: nit: let's open an issue for this instead

src/mozanalysis/experiment.py
```diff
@@ -409,10 +421,14 @@ def build_metrics_query(
...
+        WITH enrollments AS (
+            SELECT *
+            FROM {enrollments_table}
+        )
         SELECT
-            {enrollments_table}.*,
+            enrollments.*,
             {metrics_columns}
-        FROM {enrollments_table}
+        FROM enrollments
```
tdsmith, line 431: - We should write `` `{enrollments_table}` `` to make sure it's escaped, right?
- The CTE seems more complex than writing ``FROM `{enrollments_table}` AS enrollments``

tdsmith (approved):
Thanks for taking this and fixing the breakage.

Did you want to change the name of `run_query_and_fetch`?
It would be useful to specify in the docstring for that function that it's expecting a naked table name.

felixlawrence:
> Did you want to change the name of `run_query_and_fetch`?

What would you suggest?

src/mozanalysis/experiment.py
felixlawrence: > The CTE seems more complex than writing `FROM `{enrollments_table}` AS enrollments`

Yeah that's a much neater solution...

tdsmith:
I thought I'd remembered you suggesting that the `_and_fetch` was misleading but I don't insist that you change it. :)

I'd probably choose to emphasize that it can return a cached value, so something like `maybe_run_query`, but `run_query` also seems fine, or whatever you prefer.

scholtzan (approved):
lgtm

src/mozanalysis/experiment.py
felixlawrence: Huh, this appears not to work with the data source subquery in the left join - maybe I tried this originally last year and switched to the CTE accordingly. Back to the ugly version then, until BQ supports the neater one...

PR #125, opened 2021-02-22
Make enrollments tables independent of analysis windows

scholtzan:
Related to https://github.com/mozilla/jetstream/issues/455

This moves joining enrollments with the analysis windows ends and starts to `build_metrics_query`. This will allow to cache and re-use enrollments tables in jetstream.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -369,10 +369,6 @@ class Experiment:
         Returns:
             A string containing a BigQuery SQL expression.
         """
-        analysis_windows_query = self._build_analysis_windows_query(
-            time_limits.analysis_windows
-        )
-
         enrollments_query = custom_enrollments_query or self._build_enrollments_query(
             time_limits, enrollments_query_type
         )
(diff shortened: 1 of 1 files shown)


src/mozanalysis/experiment.py
```diff
@@ -369,30 +369,19 @@ def build_enrollments_query(
...
-            WITH analysis_windows AS (
-                {analysis_windows_query}
-            ),
-            raw_enrollments AS ({enrollments_query}),
+            WITH raw_enrollments AS ({enrollments_query}),
             segmented_enrollments AS ({segments_query})
 
-            SELECT
-                se.*,
-                aw.*
```
scholtzan, line 391: The `analysis_window_start` and `analysis_window_end` fields will not be part of the enrollments tables anymore? Are there any (query) dependencies that I'm missing that would break once these are gone?

scholtzan:
I did check if changing these queries would cause "resource exceeded" errors in BigQuery. I believe `bug_1671484_pref_validation_of_relpreload_performance_impact_release_82_83` was the experiment that first caused these issues which was why enrollments and metrics got separated into multiple queries. Moving the analysis window JOIN part to the metrics query didn't cause any problems for this experiment.

tdsmith (approved):
I was afraid this was gonna be really tricky so I set aside an hour to review this before looking at it, but, no, this seems obviously correct. The segments query also only depends on the span of the enrollment period, so that's reasonable, too. 🎉

scholtzan:
Is there anything I need to do for this to get published? Or is this going to happen automatically?

tdsmith:
To publish a new release you have to publish a new calver-style ("2021.2.1") tag by either creating a new release (https://github.com/mozilla/mozanalysis/releases) or a dance that looks like

```
git tag -a 2021.2.1 -m 2021.2.1
git push origin --tags
```

on an up-to-date main branch

scholtzan:
Looks like it worked

PR #128, opened 2021-04-09
Add support for Firefox for iOS experiments

scholtzan:
Required for https://github.com/mozilla/jetstream/issues/616

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -458,8 +458,6 @@ class Experiment:
         """Return SQL to query a list of enrollments and their branches"""
         if enrollments_query_type == "normandy":
             return self._build_enrollments_query_normandy(time_limits)
-        elif enrollments_query_type == "fenix-fallback":
-            return self._build_enrollments_query_fenix_baseline(time_limits)
         elif enrollments_query_type == "glean-event":
             if not self.app_id:
                 raise ValueError(
@@ -467,6 +465,10 @@ class Experiment:
                 )
             return self._build_enrollments_query_glean_baseline(
                 time_limits, self.app_id
+            )
+        elif enrollments_query_type == "fenix-fallback":
+            return self._build_enrollments_query_glean_baseline(
+                time_limits, self.app_id or "org_mozilla_firefox"
             )
         else:
             raise ValueError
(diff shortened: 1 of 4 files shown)


tdsmith:
This makes sense. I think we'd prefer to start using the Nimbus Glean events now that we have them. Instead of "ios-fallback" let's add "glean-event" which uses the [Nimbus telemetry](https://github.com/mozilla-mobile/android-components/blob/master/components/service/nimbus/docs/metrics.md). It can be an error if app_id isn't defined.

We can start depending on the Glean query type for Fenix as well once those events hit release, which I think may have happened.

src/mozanalysis/experiment.py
```diff
@@ -464,8 +458,17 @@ def _build_enrollments_query(self, time_limits, enrollments_query_type):
...
+            enrollments_query_type == "glean-event"
+            or enrollments_query_type == "fenix-fallback"
+        ):
+            if not self.app_id:
+                raise ValueError(
+                    "App ID must be defined for building Glean enrollments query"
+                )
+            return self._build_enrollments_query_glean_baseline(
+                time_limits, self.app_id
+            )
```
scholtzan, line 471: I'm seeing enrollment events for Fenix but not for iOS (yet?)

tests/test_experiment.py
```diff
@@ -374,8 +375,8 @@ def test_app_id_propagates():
     sql_lint(metrics_sql)
 
 
-def test_query_not_detectably_malformed_fenix_fallback():
```
scholtzan, line 377: I guess we don't need the fallback anymore then?

src/mozanalysis/experiment.py
```diff
@@ -464,8 +458,17 @@ def _build_enrollments_query(self, time_limits, enrollments_query_type):
         """Return SQL to query a list of enrollments and their branches"""
         if enrollments_query_type == "normandy":
             return self._build_enrollments_query_normandy(time_limits)
-        elif enrollments_query_type == "fenix-fallback":
-            return self._build_enrollments_query_fenix_baseline(time_limits)
+        elif (
+            enrollments_query_type == "glean-event"
+            or enrollments_query_type == "fenix-fallback"
```
scholtzan, line 463: Should we keep the "fenix-fallback" option around though for now for backwards compatibility? Is anyone else but Jetstream using this?

tdsmith: Let's keep it around for interactive use for now unless it's getting in your way.

src/mozanalysis/experiment.py
tdsmith: Oh interesting; let's check in with Travis on Monday then.

src/mozanalysis/experiment.py
scholtzan: We can keep it around and remove it later

scholtzan:
This should be ready for review now

src/mozanalysis/experiment.py
```diff
@@ -506,29 +510,28 @@ def _build_enrollments_query_fenix_baseline(self, time_limits):
...
-        FROM `moz-fx-data-shared-prod.{dataset}.baseline` b
-        WHERE
-            DATE(b.submission_timestamp)
+            SELECT events.client_info.client_id AS client_id,
+                mozfun.map.get_key(
+                    e.extra,
+                    'branch'
+                ) AS branch,
+                DATE(MIN(events.submission_timestamp)) AS enrollment_date
+            FROM `moz-fx-data-shared-prod.{dataset}.events` events,
```
tdsmith, line 519: I think the baseline/fallback method shouldn't change—it's intentionally using the baseline ping for cases where events aren't available. I'm not sure it will remain useful but I don't think we should change its behavior. I think there needs to be a separate method for Nimbus events.

tdsmith (changes requested):

src/mozanalysis/experiment.py
scholtzan: Added the fenix-fallback query back

src/mozanalysis/experiment.py
```diff
@@ -531,6 +532,43 @@ def _build_enrollments_query_fenix_baseline(self, time_limits):
             dataset=self.app_id or "org_mozilla_firefox",
         )
 
+    def _build_enrollments_query_glean_baseline(self, time_limits, dataset):
```
tdsmith, line 535: Let's name this `glean_event` since it doesn't use the baseline ping like the fallback method does (no idea why we adopted two different names for it)

src/mozanalysis/experiment.py
```diff
@@ -531,6 +532,43 @@ def _build_enrollments_query_fenix_baseline(self, time_limits):
...
+
+        If enrollment events are available for this experiment, then you
+        can take a better approach than this method. But in the absence
+        of enrollment events (e.g. in a Mako-based experiment, which
+        does not send enrollment events), you need to fall back to using
+        ``ping_info.experiments`` to get a list of who is in what branch
+        and when they enrolled.
+        """
+        # Try to ignore users who enrolled early - but only consider a
+        # 7 day window
```
tdsmith, line 546: I don't know why we do this for the fallback mode. In any case, we don't have to worry about it here, since the SDK will only send one enrollment event per client device.

src/mozanalysis/experiment.py
```diff
@@ -531,6 +532,43 @@ def _build_enrollments_query_fenix_baseline(self, time_limits):
...
+            UNNEST(events.events) AS e
+            WHERE
+                DATE(events.submission_timestamp)
+                BETWEEN DATE_SUB('{first_enrollment_date}', INTERVAL 7 DAY)
+                AND '{last_enrollment_date}'
+                AND e.category = "nimbus_events"
+                AND mozfun.map.get_key(e.extra, "experiment") = '{experiment_slug}'
+                AND e.name = 'enrollment'
+            GROUP BY client_id, branch
+            HAVING enrollment_date >= '{first_enrollment_date}'
```
tdsmith, line 564: in line with comment above

```suggestion
```

src/mozanalysis/experiment.py
```diff
@@ -531,6 +532,43 @@ def _build_enrollments_query_fenix_baseline(self, time_limits):
...
+                mozfun.map.get_key(
+                    e.extra,
+                    'branch'
+                ) AS branch,
+                DATE(MIN(events.submission_timestamp)) AS enrollment_date
+            FROM `moz-fx-data-shared-prod.{dataset}.events` events,
+            UNNEST(events.events) AS e
+            WHERE
+                DATE(events.submission_timestamp)
+                BETWEEN DATE_SUB('{first_enrollment_date}', INTERVAL 7 DAY)
```
tdsmith, line 558: in line with comment above

```suggestion
                BETWEEN '{first_enrollment_date}'
```

tdsmith (approved):
r+wc, thanks!

PR #129, opened 2021-05-04
Determine exposure using segments

scholtzan:
https://github.com/mozilla/jetstream/issues/649

The [proposal mentions "Exposure metrics"](https://docs.google.com/document/d/1eO-1K8j22zi-aVNCnNZICoS6nZoOv_UdcHKr0TGq4WM/edit#heading=h.yxojgorgcx1p) as one option to add exposure information to the enrollment tables. Wouldn't it be possible to reuse some of the logic we already have for segments and add an "exposed" segment to the enrollment tables? I might be missing something here.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -117,6 +117,7 @@ class Experiment:
         enrollments_query_type="normandy",
         custom_enrollments_query=None,
         segment_list=None,
+        exposure=None,
     ):
         """Return a DataFrame containing per-client metric values.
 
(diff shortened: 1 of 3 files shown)


tdsmith:
Segments are only supposed to reflect activity that happens before a user enrolls; they're not supposed to depend on user behaviour during the experiment. Right now you can't instantiate a SegmentDataSource that includes time after enrollment; it'll error. Segments also embed knowledge about a time window, and we want that to depend on the enrollment period configuration of the experiment, rather than embedding it in the segment.

So I think segments have some of the characteristics we'd like but may not be a good frame.

scholtzan:
Succeeded by https://github.com/mozilla/mozanalysis/pull/130

PR #130, opened 2021-05-05
Exposure query

scholtzan:
https://github.com/mozilla/jetstream/issues/649

Does this approach seem right?

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -117,7 +117,7 @@ class Experiment:
         enrollments_query_type="normandy",
         custom_enrollments_query=None,
         custom_exposure_query=None,
-        exposure_metric=None,
+        exposure_signal=None,
         segment_list=None,
     ):
         """Return a DataFrame containing per-client metric values.
@@ -152,6 +152,20 @@ class Experiment:
                 (client_id, branch), or else your results will be subtly
                 wrong.
 
+            custom_exposure_query (str): A full SQL query to be used in the main
+                query::
+
+                    WITH ...
+                    exposures AS ({custom_exposure_query})
+
+                If not provided, the exposure will be determined based on
+                `exposure_signal`, if provided, or Normandy and Nimbus exposure events.
+                `custom_exposure_query` takes precedence over `exposure_signal`.
+
+            exposure_signal (ExposureSignal): Optional signal definition of when a
+                client has been exposed to the experiment. If not provided,
+                the exposure will be determined based on Normandy exposure events
+                for desktop and Nimbus exposure events for Fenix and iOS.
             segment_list (list of mozanalysis.segment.Segment): The user
                 segments to study.
 
@@ -195,7 +209,7 @@ class Experiment:
             enrollments_query_type=enrollments_query_type,
             custom_enrollments_query=custom_enrollments_query,
             custom_exposure_query=custom_exposure_query,
-            exposure_metric=exposure_metric,
+            exposure_signal=exposure_signal,
             segment_list=segment_list,
         )
 
@@ -235,7 +249,7 @@ class Experiment:
         enrollments_query_type="normandy",
         custom_enrollments_query=None,
         custom_exposure_query=None,
-        exposure_metric=None,
+        exposure_signal=None,
         segment_list=None,
     ):
         """Return a TimeSeriesResult with per-client metric values.
@@ -273,10 +287,10 @@ class Experiment:
                     exposures AS ({custom_exposure_query})
 
                 If not provided, the exposure will be determined based on
-                `exposure_metric`, if provided, or Normandy and Nimbus exposure events.
-                `custom_exposure_query` takes precedence over `exposure_metric`.
+                `exposure_signal`, if provided, or Normandy and Nimbus exposure events.
+                `custom_exposure_query` takes precedence over `exposure_signal`.
 
-            exposure_metric (Metric): Optional metric definition of when a
+            exposure_signal (ExposureSignal): Optional signal definition of when a
                 client has been exposed to the experiment. If not provided,
                 the exposure will be determined based on Normandy exposure events
                 for desktop and Nimbus exposure events for Fenix and iOS.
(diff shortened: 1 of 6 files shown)


tests/test_experiment.py
```diff
@@ -446,3 +446,51 @@ def test_firefox_ios_app_id_propagation():
...
+    enrollment_sql = exp.build_enrollments_query(
+        time_limits=tl,
+        enrollments_query_type="glean-event",
+        exposure_metric=Metric(
+            name="exposure",
+            data_source=mozanalysis.metrics.fenix.baseline,
+            select_expr="metrics.counter.events_total_uri_count > 0",
+            friendly_name="URI visited exposure",
+            description="Exposed when URI visited",
+        ),
```
scholtzan, line 490: I'm wondering if it makes sense to have an `ExposureMetric` type here? By default `Metric`s are assumed to have a `select_expr` that is used in a `GROUP BY`. But in the current implementation the `select_expr` is used as a condition of when an exposure happened.

src/mozanalysis/experiment.py
```diff
@@ -367,17 +393,35 @@ def build_enrollments_query(
...
+            SELECT
+                se.*,
+                e.*
             FROM segmented_enrollments se
+            LEFT JOIN exposures e
+            USING (client_id, branch)
         """.format(
             enrollments_query=enrollments_query,
             segments_query=segments_query,
+            exposure_query=exposure_query,
```
scholtzan, line 424: This would add two columns to the enrollment tables `exposure_date` and `num_exposure_events`. I think having a single table with enrollment and exposure data is slightly more convenient for building the metrics queries rather than having separate enrollment and exposure tables.

src/mozanalysis/experiment.py
```diff
@@ -563,6 +626,74 @@ def _build_enrollments_query_glean_event(self, time_limits, dataset):
...
+            WHERE
+                event_category = 'normandy'
+                AND (event_method = 'exposure' AND event_method = 'expose')
+                AND submission_date
+                    BETWEEN '{first_enrollment_date}' AND '{last_enrollment_date}'
+                AND event_string_value = '{experiment_slug}'
+        ) e
+        ON re.client_id = e.client_id AND
+            re.branch = e.branch AND
+            e.submission_date >= re.enrollment_date
```
scholtzan, line 654: This wasn't explicitly mentioned in the proposal, but do we only consider the exposures after the enrollment or should exposures during the enrollment window (which I guess could happen before an enrollment event was received) be considered?

tdsmith: After enrollment sounds right to me. If the exposure event isn't experiment-specific, then we don't want to include exposure that didn't reflect the experiment content.

src/mozanalysis/experiment.py
```diff
@@ -260,6 +266,14 @@ def get_time_series_data(
                 (client_id, branch), or else your results will be subtly
                 wrong.
 
+            custom_exposure_query (str): A full SQL query to be used in the main
+                query::
+
+                    WITH ...
+                    exposures AS ({custom_exposure_query})
```
tdsmith, line 273: Can you document what happens if it isn't specified? How do I use the Nimbus exposure event? Ditto for exposure_metric.

src/mozanalysis/experiment.py
```diff
@@ -367,17 +393,35 @@ def build_enrollments_query(
             time_limits, enrollments_query_type
         )
 
+        if exposure_metric:
+            exposure_query = (
+                custom_exposure_query
```
tdsmith, line 398: Is this branch possible?

src/mozanalysis/experiment.py
```diff
@@ -563,6 +626,74 @@ def _build_enrollments_query_glean_event(self, time_limits, dataset):
...
+                FROM
+                    `moz-fx-data-shared-prod.{dataset}.events` events,
+                    UNNEST(events.events) AS e
+                WHERE
+                    DATE(events.submission_timestamp)
+                    BETWEEN '{first_enrollment_date}' AND '{last_enrollment_date}'
+                    AND e.category = "nimbus_events"
+                    AND mozfun.map.get_key(e.extra, "experiment") = '{experiment_slug}'
+                    AND (e.name = 'expose' OR e.name = 'exposure')
+            ) e
```
tdsmith, line 685: nit: a couple of different things are aliased to `e` here; can they take different names?

tests/test_experiment.py
tdsmith: If they have different semantics I agree the types should be different, but could they have the same semantics? It seems like we could group by client_id and branch (sort of unnecessarily) and have the same aggregation semantics. Is that a good idea?

tdsmith (approved):

src/mozanalysis/experiment.py
scholtzan: Yeah. It is possible to specify both `exposure_metric` and `custom_exposure_query` in `get_time_series_data()`. However, `custom_exposure_query` will take precedence

tests/test_experiment.py
scholtzan: Kind of. However, I think a separate type makes sense here to make the distinction between normal metrics and exposure metrics clearer. Otherwise, it might not be entirely clear that most of the pre-defined metrics, like `metrics.desktop.baseline_ping_count`, don't work correctly here although they have the right type. `select_expr` still expects some kind of condition to determine when an exposure happened vs. just an aggregation of metric values.

tdsmith: Sounds good! :bikeshed: but maybe this isn't a `Metric` at all then; how about a `ExposureSignal`? It would be satisfying if we could implement the Nimbus exposure event in terms of whatever the abstraction we use for these events is.

scholtzan: Calling it `ExposureSignal` sounds good 👍

src/mozanalysis/exposure.py
```diff
@@ -0,0 +1,69 @@
...
+
+from typing import Optional
+
+import attr
+
+from mozanalysis.metrics import DataSource
+
+
+@attr.s(frozen=True, slots=True)
+class ExposureSignal:
```
scholtzan, line 13: @tdsmith I added `ExposureSignal` here, if you want to take another look

PR #134, opened 2021-06-16
Rename search_clients_daily

tdsmith:


src/mozanalysis/metrics/desktop.py
```diff
@@ -11,18 +11,22 @@
...
-#: .. _`search_clients_daily`: https://docs.telemetry.mozilla.org/datasets/
-#:    search/search_clients_daily/reference.html
-search_clients_daily = DataSource(
-    name="search_clients_daily",
-    from_expr="`moz-fx-data-shared-prod.search.search_clients_daily`",
+#: .. _`search_clients_engines_sources_daily`: https://docs.telemetry.mozilla.org/
+#:    datasets/search/search_clients_engines_sources_daily/reference.html
+search_clients_engines_sources_daily = DataSource(
+    name="search_clients_engines_sources_daily",
+    from_expr="`moz-fx-data-shared-prod.search.search_clients_engines_sources_daily`",
```
jklukas, line 22: Note that all user-facing views are published in both shared-prod and mozdata. Preferred style today would be to use mozdata and drop the backticks here.

tdsmith: Thanks for the callout!

src/mozanalysis/metrics/desktop.py
```diff
@@ -61,7 +65,7 @@
                     *,
                     DATE(submission_timestamp) AS submission_date,
                     environment.experiments
-                FROM `moz-fx-data-shared-prod`.telemetry_stable.main_v4
+                FROM mozdata.telemetry_stable.main_v4
```
scholtzan, line 68: This needs to be `moz-fx-data-shared-prod`. `mozdata` doesn't have access to stable or derived tables

```suggestion
                FROM `moz-fx-data-shared-prod`.telemetry_stable.main_v4
```

scholtzan (approved):
r+wc

src/mozanalysis/metrics/desktop.py
tdsmith: thanks for the catch!

PR #136, opened 2021-06-23
Add autodocs for segments

tdsmith:
Also autodoc segments

### DIFF
--- a/src/mozanalysis/segments/desktop.py
+++ b/src/mozanalysis/segments/desktop.py
@@ -6,6 +6,7 @@ from textwrap import dedent
 from mozanalysis.metrics import agg_any
 from mozanalysis.segments import Segment, SegmentDataSource
 
+#: SegmentDataSource: The clients_last_seen table.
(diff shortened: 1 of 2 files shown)


src/mozanalysis/segments/desktop.py
```diff
@@ -6,13 +6,15 @@
 from mozanalysis.metrics import agg_any
 from mozanalysis.segments import Segment, SegmentDataSource
 
+#: SegmentDataSource: The clients_last_seen table.
 clients_last_seen = SegmentDataSource(
     name="clients_last_seen",
     from_expr="`moz-fx-data-shared-prod.telemetry.clients_last_seen`",
```
scholtzan, line 12: At some point we should probably switch to `mozdata`, I guess

scholtzan (approved):

src/mozanalysis/segments/desktop.py
tdsmith: No time like the present!

PR #147, opened 2021-12-07
Specify __all__ for metrics and segments module

scholtzan:
Required for https://github.com/mozilla/jetstream/pull/971


kik-kik (approved):
LGTM 👍 

But since we're repeating the logic maybe it would be worth defining it at the parent level as a function and just calling it here to avoid code duplication?

PR #151, opened 2022-07-07
Add QCDOU to Desktop Metrics

rebecca-burwei:


rebecca-burwei:
I am following [this](https://github.com/mozilla/jetstream-config/blob/main/defaults/firefox_desktop.toml#L78), but why do we refer [this](https://github.com/mozilla/jetstream-config/blob/main/defaults/firefox_desktop.toml#L78) as cumulative and [this](https://github.com/mozilla/mozanalysis/blob/f142c2132fcffae3aaab1c79aaf5117df2ad3f77/src/mozanalysis/metrics/desktop.py#L347) as not-cumulative? @danielkberry

scholtzan (approved):
lgtm. I'll let Daniel review the metric definition.

Once this has been merged we'll have to deploy a new version (by adding a new tag) to make this available in jetstream + update jetstream to use the new mozanalysis version.

scholtzan:
FYI the goal is at some point to move all the metric definitions to jetstream-config and have mozanalysis import and use them from there. However, since the work to make that happen will probably take a few more weeks/months, this is fine.

The work for that is tracked in https://github.com/mozilla/jetstream/issues/1131

danielkberry (approved):
Agree with Anna's comments. We're going to be centralizing all of these somewhere, eventually, so we'll be moving this around again, but for now this should be ok.

scholtzan:
I'll merge this now and cut a new release.

PR #152, opened 2022-08-04
Power calculation

m-d-bowerman:
These changes add functionality to Mozanalysis to pull historical data for use in experiment sample size calculation. The changes include a class to define historical data sources, a class to define filters on those data sources, and a class to construct target and metrics queries to pull the data, along with method to perform sample size calculation. 

1. Adds historical_targets module
- This includes the TargetDataSource and Target classes, that hold information about the targets for the experiment sizing
- adds desktop.py that holds TargetDataSource and Target classes transcribed from the desktop Segments in mozanalysis
2. Adds sizing.py
- This includes the HistoricalTarget class that is used to construct the target and metrics queries
- The class includes a test use case in the comments
- The method get_single_window_data returns a DataFrame for analyses using historical data
3. Adds build_query_targets to Metric class
- Method is based off the existing build_query method in Metric class
- Joins to target table instead of enrollment, and references to experiment slug and branch columns in build_query result were removed
4. Adds sample_size_calc function to frequentist_stats module
- This function does sample size calculation using the results from get_single_window_data in HistoricalTarget


src/mozanalysis/historical_targets/__init__.py
```diff
@@ -0,0 +1,109 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from typing import Optional
+import datetime
+
+import attr
+
+@attr.s(frozen=True, slots=True)
+class TargetDataSource:
```
rebecca-burwei, line 10: Is it necessary to copy all of this code over from [metrics.DataSource](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/metrics/__init__.py#L12), or can we inherit / reuse some of it? It seems like a lot of code to have duplicated.

src/mozanalysis/historical_targets/desktop.py
```diff
@@ -0,0 +1,80 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from textwrap import dedent
+
+from mozanalysis.metrics import agg_any
+from mozanalysis.historical_targets import TargetDataSource, Target
+
+#: SegmentDataSource: The clients_last_seen table.
+clients_last_seen = TargetDataSource(
```
rebecca-burwei, line 10: Can we reuse [this](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/segments/desktop.py)? It would be less confusing and more maintainable to not have 2 sources of truth for client segments.

rebecca-burwei (changes requested):
I'd like to see us duplicate less code and try a bit more to repurpose or inherit from the existing mozanalysis classes.

src/mozanalysis/historical_targets/__init__.py
m-d-bowerman: There are some slight differences between the metrics DataSource and here (similar to differences between SegmentDataSource and DataSource). The closest analog would be Segments, but I defined a new class for Targets following that pattern

src/mozanalysis/historical_targets/desktop.py
m-d-bowerman: I created a new source here because Segments require aggregating SQL functions to work, and I saw possible Targets that would not need them (e.g. locale, language, etc)

rebecca-burwei: Could you say more about the case where an aggregation is not needed? For locale, users can change their locale, so I would think something like `agg_any("locale LIKE 'en*'")` would be needed to target all clients that had set their locale to english at any point during a certain time period.

m-d-bowerman: Ah, gotcha, that's my unfamiliarity with the data coming through. I'll make the change to reuse the SegmentDataSource objects, thanks!

src/mozanalysis/historical_targets/__init__.py
rebecca-burwei: ah okay - that makes sense.

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,358 @@
...
+
+        ht = HistoricalTarget(
+            experiment_name='my_test_name',
+            start_date='2021-01-01',
+            num_dates_enrollment=8,
+            print_queries=True
+        )
+
+        # Run the query and get the results as a DataFrame
+        res = experiment.get_single_window_data(
```
danielkberry, line 43: Should this be `res = ht.get_single_window_data(`?

src/mozanalysis/historical_targets/desktop.py
rebecca-burwei: no problem :)

src/mozanalysis/sizing.py
m-d-bowerman: You're right, it should be. I'll fix that, thanks!

src/mozanalysis/historical_targets/__init__.py
m-d-bowerman: Okay, I've moved the query building from the Target and TargetDataSource to the SegmentDataSource class, so the HistoricalTarget will now take Segment objects

src/mozanalysis/historical_targets/desktop.py
m-d-bowerman: Same as above, made the requested change

src/mozanalysis/sizing.py
m-d-bowerman: Fixed!

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,383 @@
...
+        bq_context = BigQueryContext(
+            dataset_id='your-dataset-id',  # e.g. mine's mbowerman
+            project_id='moz-fx-data-bq-data-science'  # this is the default anyway
+        )
+
+        ht = HistoricalTarget(
+            experiment_name='my_test_name',
+            start_date='2021-01-01',
+            num_dates_enrollment=8,
+            print_queries=True
```
danielkberry, line 38: I get a `TypeError` when running this line, looks like `print_queries` [isn't listed as a parameter](https://github.com/mozilla/mozanalysis/pull/152/files#diff-9ac2a7afe5f1013fb39513b6c52d0481e5b4210234d30ddc97f490b7aa410f8bR101-R106)
>TypeError: __init__() got an unexpected keyword argument 'print_queries'

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,383 @@
...
+            metric_list = [
+                active_hours,
+                uri_count
+            ],
+            last_date_full_data='2021-01-30',
+            analysis_start_days=0,
+            analysis_length_days=21,
+            target_list = [
+                new_unique_profiles,
+                new_or_resurrected_v3
```
danielkberry, line 53: This object [needs to be imported](https://github.com/mozilla/mozanalysis/pull/152/files#diff-9ac2a7afe5f1013fb39513b6c52d0481e5b4210234d30ddc97f490b7aa410f8bR27) along with `new_unique_profiles`

m-d-bowerman: Fixed, thanks!

src/mozanalysis/sizing.py
m-d-bowerman: Fixed

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,383 @@
...
+    metrics_sql = attr.ib(default='')
+
+    def get_single_window_data(
+        self,
+        bq_context,
+        metric_list,
+        last_date_full_data,
+        analysis_start_days,
+        analysis_length_days,
+        target_list=None,
```
danielkberry, line 115: Given the move to `Segment` objects, should this be `segment_list`?

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,383 @@
...
+                all data collected after a certain date (e.g. when the
+                experiment recipe was deactivated), then do that here.
+            analysis_start_days (int): the start of the analysis window,
+                measured in 'days since the client enrolled'. We ignore data
+                collected outside this analysis window.
+            analysis_length_days (int): the length of the analysis window,
+                measured in days.
+            target_list (list of mozanalysis.segments.Segment): The targets
+                that define clients to be included in the analysis, based on
+                inclusion in a defined user segment.
```
danielkberry, line 139: Can we add something here to explain how multiple `Segment`s will be and'd together?

src/mozanalysis/sizing.py
m-d-bowerman: It could, but I kept this naming to note the distinction between how Segment objects are used here, where they define the exact client IDs that are included, versus how they're used in Experiment, where they add a binary column indicating segment inclusion

danielkberry: Ahh, makes sense. 👍

src/mozanalysis/sizing.py
```diff
@@ -0,0 +1,383 @@
...
+            factor '7' removes weekly seasonality, and the ``+1`` accounts
+            for the fact that enrollment typically starts a few hours
+            before UTC midnight.
+        analysis_length (int, optional): Number of days to include for analysis
+        targets_sql (str, optional): The SQL query that was executed to create
+            the targets
+            table; only defined after get_single_window_data method is called.
+        metrics_sql (str, optional): The SQL query that was executed to create the final
+            results table; only defined after get_single_wondow_data method is
+            called.
```
danielkberry, line 77: Are these actually parameters that someone might pass in? Or are they just attributes that (sometimes) exist in the object? If it's the latter, what if we instead just made them private attributes and provided a getter method:
```python
def get_targets_sql(self):
    if not self._targets_sql:
        raise ValueError("Target SQL not available, must call `get_single_window_data` first") # or something
    return self._targets_sql
```

src/mozanalysis/experiment.py
```diff
@@ -1152,13 +1152,87 @@ def get(self, bq_context, analysis_window):
...
+        )
+
+        return bq_context.run_query(
+            """
+            SELECT * EXCEPT (client_id)
+            FROM {full_table_name}
+            """.format(
+                full_table_name=self.fully_qualified_table_name
+            )
+        ).to_dataframe()
```
scholtzan, line 1181: This could also be written as something like, if it is fine to keep the `client_id`:

```suggestion
table = bq_context.client.get_table(full_table_name)
return bq_context.client.list_rows(table).to_dataframe()
```

This might be a little more performant

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -0,0 +1,283 @@
...
+
+    res_mean = res.get_aggregated_data(
+        bq_context=bq_context,
+        metric_list=metric_list,
+        aggregate_function="AVG"
+    )
+
+    res_mean.sort_values(by="analysis_window_start", ascending=True, inplace=True)
+    effect_size = {
+        m.name: res_mean[m.name].diff().quantile(q=quantile) for m in metric_list
```
rebecca-burwei, line 210:
```suggestion
        m.name: res_mean[m.name].diff().abs().quantile(q=quantile) for m in metric_list
```
Could we take the absolute value of differences?

m-d-bowerman: Yup! I made that change in the most recent commit

rebecca-burwei:
@m-d-bowerman I'm getting some errors today that I didn't have yesterday. Here are some relevant snippets:

Code & error message:
image

Part of the generated SQL:
image

rebecca-burwei (approved):
this approval is long overdue. :) duplicate code was removed

danielkberry:
I'm working on merging my changes (adding types and support for glean data), then can provide a review

m-d-bowerman:
> @m-d-bowerman I'm getting some errors today that I didn't have yesterday. Here are some relevant snippets:
> 
> Code & error message: image
> 
> Part of the generated SQL: image

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+    metric_names = [m.name for m in metrics_list]
+    results = {}
+    for col in metric_names:
+        sample_size = _get_sample_size_col(col)
+        pop_percent = 100.*(sample_size/len(df))
+        results[col] = {
+            "sample_size": sample_size,
+            "population_percent": pop_percent
```
rebecca-burwei, line 70:
```suggestion
            "population_percent": pop_percent,
             "population": len(df)
```
Could we include the population as well? It's helpful to have this number as a sanity check.

m-d-bowerman: Added!

rebecca-burwei: thanks!

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -0,0 +1,313 @@
...
+        bq_context=bq_context,
+        metric_list=metric_list,
+        aggregate_function="STDDEV"
+    )
+    std = {m.name: res_std[m.name].quantile(q=quantile) for m in metric_list}
+
+    size_dict = {
+        m.name: {
+            "effect_size": effect_size[m.name],
+            "std_dev": std[m.name],
```
rebecca-burwei, line 239: @m-d-bowerman could we add to this output the analysis windows that the effect size and standard deviation correspond to?

m-d-bowerman: Just to make sure I understand, you want it to return the time period that corresponds to the 90th percentile for each statistic?

rebecca-burwei: yes - either the exact dates of the time period, or even a window identifier would be fine.

m-d-bowerman: The change to add the analysis window to the results dictionary is included in the latest commit

rebecca-burwei: thank you!

src/mozanalysis/metrics/__init__.py
```diff
@@ -177,7 +187,66 @@ def build_query(
...
+        FROM targets t
+            LEFT JOIN {from_expr} ds
+                ON ds.{client_id} = t.client_id
+                {date_clause}
+        GROUP BY
+            t.client_id,
+            t.enrollment_date,
+            t.analysis_window_start,
+            t.analysis_window_end""".format(
+            client_id=self.client_id_column,
```
scholtzan, line 222:
```suggestion
            client_id=self.client_id_column or "client_id",
```

src/mozanalysis/metrics/__init__.py
```diff
@@ -177,7 +187,66 @@ def build_query(
...
+                    se=m.select_expr.format(experiment_name=experiment_name), n=m.name
+                )
+                for m in metric_list
+            ),
+            date_clause="""
+        AND ds.{submission_date} BETWEEN '{fddr}' AND '{lddr}'
+        AND ds.{submission_date} BETWEEN
+            DATE_ADD(t.enrollment_date, interval t.analysis_window_start day) AND
+            DATE_ADD(t.enrollment_date, interval t.analysis_window_end day)""".format(
+                submission_date=self.submission_date_column,
```
scholtzan, line 235:
```suggestion
                submission_date=self.submission_date_column or "submission_date",
```

src/mozanalysis/metrics/__init__.py
```diff
@@ -177,7 +187,66 @@ def build_query(
...
+                fddr=time_limits.first_date_data_required,
+                lddr=time_limits.last_date_data_required,
+            )
+            if not continuous_enrollment
+            else """AND ds.{submission_date} BETWEEN
+            t.enrollment_date AND
+            DATE_ADD(t.enrollment_date, interval {analysis_length} day)
+            """.format(
+                submission_date=self.submission_date_column,
```
scholtzan, line 244:
```suggestion
                submission_date=self.submission_date_column or "submission_date",
```

src/mozanalysis/segments/__init__.py
```diff
@@ -109,6 +109,44 @@ def build_query(
...
+        FROM (SELECT {client_id},
+                MIN({submission_date}) as target_first_date,
+                MAX({submission_date}) as target_last_date,
+                {target}
+            FROM {from_expr}
+            WHERE {submission_date} BETWEEN '{fddr}' AND '{lddr}'
+            GROUP BY {client_id})
+        WHERE {target_name}""".format(
+            client_id=self.client_id_column,
+            submission_date=self.submission_date_column,
```
scholtzan, line 142:
```suggestion
            client_id=self.client_id_column or "client_id",
            submission_date=self.submission_date_column or "submission_date",
```

scholtzan (commented):
Something I just came across while working on the metrics parser library and that we talked about as part of https://github.com/mozilla/metric-config-parser/pull/42

The [`client_id_column` and `submission_date` fields](https://github.com/mozilla/metric-config-parser/blob/91ee9f57e795719536f99b7f12add340a942d159/metric_config_parser/data_source.py#L122-L123) are defined as optional so they can (and will be null) for some of the data source definitions. So it is probably good to check for this null condition here and cast to a default value. 

I'll need to remove the default values from the metric parser library in a newer version since they are causing some issues when overriding a few fields of an existing data source definition. In those cases the `client_id_column` and `submission_date` fields will _always_ get overridden, which is not always desired when the original definition should be kept.

rebecca-burwei:
**Power calculator and Nimbus do not agree**

1. [Here](https://experimenter.services.mozilla.com/nimbus/firefox-suggest-enhanced-exposure-phase-1-extended) is a completed Nimbus experiment that targeted 10% of its population. It enrolled ~1.32M clients, implying that the population size was 13.2M.
2. [Here](https://colab.research.google.com/drive/1tQewNmxKBxqJoklYbShx5GvAOIbw0o_U#scrollTo=ZSqHq5GmWlJG) is an example of me using the power calculator to create a HistoricalTarget for that experiment. The power calculator tells me there are 15.1M clients in the population.

@danielkberry fyi

m-d-bowerman:
> 
Thanks Rebecca! I'll compare the queries for both and try to figure out what's going on.

PR #153, opened 2022-08-08
Pull in configurations from jetstream-config

scholtzan:
With this change mozanalysis will use the new [jetstream-config-parser](https://github.com/mozilla/jetstream-config-parser) to pull in configurations from jetstream-config. The changes introduced here shouldn't break any existing code.

I've already migrated existing metric, data source and segment definitions to https://github.com/mozilla/jetstream-config/tree/main/definitions, but it would probably be best to validate them.

### DIFF
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -19,7 +19,6 @@ We take the per-notebook daily trudge::
 
 Then we import the necessary classes for getting the data, and for analysing the data, and for interacting with BigQuery::
 
-    import mozanalysis.metrics.desktop as mmd
     import mozanalysis.bayesian_stats.binary as mabsbin
     from mozanalysis.experiment import Experiment
     from mozanalysis.bq import BigQueryContext
@@ -45,7 +44,8 @@ We start by instantiating our :class:`mozanalysis.experiment.Experiment` object:
     exp = Experiment(
         experiment_slug='pref-fingerprinting-protections-retention-study-release-70',
         start_date='2019-10-29',
-        num_dates_enrollment=8
+        num_dates_enrollment=8,
+        app_name="firefox_desktop"
     )
 
 ``start_date`` is the ``submission_date`` of the first enrollment (``submission_date`` is in UTC). If you intended to study one week's worth of enrollments, then set ``num_dates_enrollment=8``: Normandy experiments typically go live in the evening UTC-time, so 8 days of data is a better approximation than 7.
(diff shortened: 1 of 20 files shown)


docs/guide.rst
```diff
@@ -45,7 +44,8 @@ We start by instantiating our :class:`mozanalysis.experiment.Experiment` object:
     exp = Experiment(
         experiment_slug='pref-fingerprinting-protections-retention-study-release-70',
         start_date='2019-10-29',
-        num_dates_enrollment=8
+        num_dates_enrollment=8,
+        app_name="firefox_desktop"
```
scholtzan, line 48: We'll need to pass the `app_name` into Experiment so we can choose the correct metric based on the platform the experiment is run on

src/mozanalysis/metrics/fenix.py
```diff
@@ -2,112 +2,29 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from mozanalysis.metrics import DataSource, Metric, agg_sum
+# These metric definitions are deprected.
```
mikewilli, line 5:
```suggestion
# These metric definitions are deprecated.
```

src/mozanalysis/experiment.py
```diff
@@ -137,7 +158,7 @@ def get_single_window_data(
 
         Args:
             bq_context (BigQueryContext): BigQuery configuration and client.
-            metric_list (list of mozanalysis.metric.Metric): The metrics
+            metric_list (list of mozanalysis.metric.Metric or str): The metrics
```
mikewilli, line 161: Just confirming I understand correctly -- use of `mozanalysis.metric.Metric` here is deprecated, and we expect to use `str` (metric slug) going forward?

src/mozanalysis/metrics/desktop.py
```diff
@@ -1,404 +1,84 @@
...
+    "search_clients_engines_sources_daily", "firefox_desktop"
 )
 
-#: DataSource: The events table.
-events = DataSource(
-    name="events",
-    from_expr="mozdata.telemetry.events",
-    experiments_column_type="native",
+search_clients_daily = ConfigLoader.get_data_source(
+    "search_clients_engines_sources_daily", "firefox_desktop"
```
mikewilli, line 18: Any reason to make another call instead of doing what we had before (below)?
```suggestion
search_clients_daily = search_clients_engines_sources_daily
```

mikewilli (approved):
Looks good to me!

src/mozanalysis/metrics/__init__.py
```diff
@@ -282,16 +282,19 @@ class Metric:
 
 def agg_sum(select_expr):
     """Return a SQL fragment for the sum over the data, with 0-filled nulls."""
+    print("The use of mozanalysis.metrics.agg_sum() is deprecated")
```
danielkberry, line 285: Any thoughts on raising Warnings instead of prints?

danielkberry (approved):
This is awesome

src/mozanalysis/experiment.py
scholtzan: That is mostly correct. `Metric` can still be used if, for example, a DS defines their own custom metric in a Jupyter notebook that uses mozanalysis. But for referencing existing metrics `str` is recommended

src/mozanalysis/metrics/desktop.py
scholtzan: Good catch, we can keep what we had before

src/mozanalysis/metrics/__init__.py
scholtzan: I saw that mozanalysis has been using `print`s to log stuff so I figured to be consistent I do the same here. But maybe we should just use `logging` in the other places as well

src/mozanalysis/bq.py
```diff
@@ -63,11 +64,11 @@ def run_query(self, sql, results_table=None):
                     )
                 ),
             ).result()
-            print("Saved into", results_table)
+            logging.info("Saved into", results_table)
```
mikewilli, line 67: Is there a reason to use `logging.info()` instead of instantiating a logger with getLogger and then using `logger.info()`?

scholtzan: yeah, getLogger probably makes more sense here

PR #155, opened 2022-11-01
fix typo in exposures query

mikewilli:
I am guessing that at some point we switched from `exposure` to `expose` as the `event_method` (or maybe were just leaving open both possibilities) and that this was meant to be `OR` to account for either option.


scholtzan (approved):
typo indeed

PR #163, opened 2023-01-20
Clarifies sample size results as per branch

m-d-bowerman:
Updates naming convention for results DataFrames for experiment sizing to denote sizes are per branch in the study.


danielkberry (approved):
LGTM

PR #164, opened 2023-02-08
Mbowerman sizing results rename

m-d-bowerman:
- Renames the results for sample size calc to make explicit that sizes are per branch of a study.
- Fixes joins for creating targets to use INNER rather than LEFT joins.

### DIFF
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -2,16 +2,15 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import attr
-from typing import List, Union, Dict, Callable, Optional
-from datetime import datetime, timedelta
+from typing import List, Union, Dict, Optional
+from datetime import datetime
 
 from mozanalysis.bq import BigQueryContext
 from mozanalysis.experiment import TimeSeriesResult
 from mozanalysis.metrics import Metric
 from mozanalysis.segments import Segment
 from mozanalysis.sizing import HistoricalTarget
-from mozanalysis.utils import add_days, get_time_intervals
+from mozanalysis.utils import get_time_intervals
 
 from scipy.stats import norm
 from math import pi
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -142,7 +142,7 @@ def difference_of_proportions_sample_size_calc(
         results[col] = {
             "sample_size_per_branch": sample_size,
             "population_percent_per_branch": pop_percent,
-            "population_size": len(df),
+            "number_of_clients_targeted": len(df),
         }
     return results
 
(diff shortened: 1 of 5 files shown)


src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -391,7 +398,86 @@ def _get_sample_size_col(col):
...
+        num_dates_enrollment=max_enrollment_days,
+        experiment_name=experiment_name,
+        app_id=app_id,
+    )
+
+    df = ht.get_single_window_data(
+        bq_context=bq_context, metric_list=metric_list, target_list=target_list
+    )
+
+    def _get_time_intervals(start_date: Union[str, datetime], interval: int):
```
danielkberry, line 437: I'm a fan of defining functions at the top level (i.e., not nested within other functions) because that way they can be independently unit tested. Especially for a simple function like this that doesn't need any extra state.

danielkberry: Also, can we add type hinting for the return type? From reading the code I see it's returning a `Dict[int, datetime]`, but that would be nice in the signature. Also maybe a doc string about what it means? What does the int key mean here?

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -391,7 +398,86 @@ def _get_sample_size_col(col):
...
+            }
+
+        return final_res
+
+    results_dict = {}
+    for m in metric_list:
+        results_dict[m.name] = []
+
+    for i in range(len(date_intervals)):
+        res = _for_interval_sample_size_calculation(i)
```
danielkberry, line 475: If I were writing this, I'd change the function signature from an `int` (which represents an index into an external map of date intervals), to the date interval itself. Like:

```python
for date_interval in date_intervals:
    res = _for_interval_sample_size_calculation(date_interval)
```

Because then above you don't need need to index into the date_intervals dict at all, which limits the sharing of state between this function and the outer scope.

danielkberry: Actually, upon further reading, it doesn't seem like we really need to key for the date_intervals dict, so what if instead we went with an array of end dates. So `_get_time_intervals` could be renamed as `_get_end_dates` and return a `List[date]`, this loop could be rewritten as:
```python
end_dates = _get_end_dates(start_date, variable_window_length)

for end_date in end_dates:
    res = _for_interval_sample_size_calculation(end_date)

```

danielkberry (commented):
Overall, I think the expanding intervals functionality is a much needed one. I added a few comments, but the gist is: I think there's a bit of unnecessary complexity in keeping a dict of end dates (which are being called time intervals). I don't think we need to use a dict and we might be able to simplify and increase readability if instead we used a list of end dates. 

Before I approve/reject, is my reading of the code correct in that we're primarily dealing with end dates and not arbitrary intervals?

src/mozanalysis/frequentist_stats/sample_size.py
m-d-bowerman: Ah, agreed. The dictionary with end dates is an artifact of a previous way I was doing this, where the intervals do matter. I'll change it to a list and disregard the keys.

src/mozanalysis/frequentist_stats/sample_size.py
m-d-bowerman: Yeah, that makes sense. I'll move this function into util.py and add type hinting for the return.

danielkberry (approved):

PR #169, opened 2023-02-16
Prefilter Data Sources using experiment annotation where available

danielkberry:


danielkberry:
This builds off of the ideas that @scholtzan came up with for the `abouthome-startup-cache` experiment, namely: we can pre-filter data sources that have experiment annotations down to only data from the appropriate experiment, thus reducing the amount of work necessary when joining data sources together.

scholtzan (approved):

danielkberry:
@scholtzan What's the procedure from here? Can I just merge this or should we test it somehow?

scholtzan:
In an ideal world we'd add a test for this in https://github.com/mozilla/mozanalysis/blob/main/tests/test_metrics.py
But in this reality, we've usually been neglecting tests a bit. So merging as is is probably fine

PR #175, opened 2023-03-27
Remove parens in metric query

scholtzan:
This was causing `Syntax error: Expected keyword JOIN but got ")"`


danielkberry (approved):
Ahh, good catch

PR #177, opened 2023-04-03
Improve the performance of bayesian bootstrap by 10x

jrmuizel:
The vast majority of the current time is in:
np.random.RandomState(unique_seed)

Previously, when generation of the sample weights was parallelized it made sense to reseed for each sampling so that it was deterministic. However, that parallelization was removed in e33914d7fa42bfa1f305d14ee7b96c3a39643abb

With that gone we can drop the reseeding entirely which makes 16 or so 100 sample calls to compare_branches go from 50seconds down to 5.

### DIFF
--- a/src/mozanalysis/bayesian_stats/bayesian_bootstrap.py
+++ b/src/mozanalysis/bayesian_stats/bayesian_bootstrap.py
@@ -294,13 +294,10 @@ def get_bootstrap_samples(
     if seed_start is None:
         seed_start = np.random.randint(np.iinfo(np.uint32).max)
 
-    # Deterministic "randomness" requires careful state handling :(
-    # Need to ensure every call has a unique, deterministic seed.
-    seed_range = range(seed_start, seed_start + num_samples)
-
+    random_state = np.random.RandomState(seed_start)
     summary_stat_samples = [
-        _resample_and_agg_once(data_values, data_counts, stat_fn, unique_seed)
-        for unique_seed in seed_range
+        _resample_and_agg_once(data_values, data_counts, stat_fn, random_state)
+        for _ in range(0, num_samples)
     ]
 
     summary_df = pd.DataFrame(summary_stat_samples)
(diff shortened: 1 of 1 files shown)


danielkberry:
Hi there, thanks for making this PR. I [tested the change](https://colab.research.google.com/drive/1CaD23pN7kh6N8QU8m20nCmbp77Cx7zQe#scrollTo=yDTBbIvmEQDF) under conditions like we might encounter in normal use of this code. This test simulated an experiment and measured the timing of `compare_branches` under both the old code and the proposed change here (as described in the 1st comment). 

As we can see from the timings in the notebook, the timings are not statistically different from each other, so this method does not appear to provide any speedup.

danielkberry (changes requested):
Please demonstrate that the code changes improve performance

jrmuizel:
Your notebook doesn't actually use the new code. If I fix that with:
```python
mozanalysis.bayesian_stats.bayesian_bootstrap.get_bootstrap_samples = get_bootstrap_samples
mozanalysis.bayesian_stats.bayesian_bootstrap._resample_and_agg_once = _resample_and_agg_once
```
 I get:

Before:
```
%%timeit
compare_branches(data, 'metric_a')
11.9 s ± 86 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

```
%%timeit
compare_branches(data, 'metric_b')
5.78 s ± 2.11 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After:
```
%%timeit
compare_branches(data, 'metric_a')
6.53 s ± 599 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
```
%%timeit
compare_branches(data, 'metric_b')
358 ms ± 84.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Here's an updated notebook: https://colab.research.google.com/drive/1p5Zl6vnmwVG8daoF4LTrNHF844zqFcCG?usp=sharing

PR #182, opened 2023-05-11
Update metrics data source experiment annotation filter

danielkberry:


src/mozanalysis/metrics/__init__.py
```diff
@@ -109,7 +109,7 @@ def from_expr_for(
                 WHERE {slug_expr}
             )""".format(
                 from_expr=from_expr,
-                slug_expr=self.experiments_column_expr_base.format(
+                slug_expr=self.experiments_column_expr.format(
                     experiment_slug=experiment_slug
```
scholtzan, line 113: something like this is missing here
```suggestion
experiment_slug=experiment_slug,
submission_date=effective_dataset.submission_date_column
```

scholtzan (approved):

PR #184, opened 2023-05-15
Revert changes from #169

scholtzan:
Going back to pre https://github.com/mozilla/mozanalysis/pull/169 so that we don't unenroll clients during analysis

### DIFF
--- a/src/mozanalysis/metrics/__init__.py
+++ b/src/mozanalysis/metrics/__init__.py
@@ -79,9 +79,7 @@ class DataSource:
     def _check_default_dataset_provided_if_needed(self, attribute, value):
         self.from_expr_for(None)
 
-    def from_expr_for(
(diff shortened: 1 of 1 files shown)


src/mozanalysis/metrics/__init__.py
```diff
@@ -97,65 +95,40 @@ def from_expr_for(
...
-        from_expr = self._from_expr.format(dataset=effective_dataset)
-
-        if (self.experiments_column_type is None) or (experiment_slug is None):
-            return from_expr
-
-        else:
-            return """(
-                SELECT *
-                FROM {from_expr} ds
-                WHERE {slug_expr}
```
scholtzan, line 109: This part was introducing the main problem of removing clients from the analysis if they unenroll during the experiment

mikewilli (approved):
Underscores the need for some more testing, but looks good.

PR #185, opened 2023-07-15
[POC | not ready for merge] Speed up bootstrap using multiprocessing

rebecca-burwei:
Trying out some things I learned at SciPy.

Locally, I'm seeing a 4-5x speed-up on the tests. I wonder what the speed-up would be in our production environments. 

I understand we're introducing a bit more non-determinism here, but I bet it's possible to make the difference small enough.

If this works, we should implement it for the Bayesian bootstrap too.

Other tips:
- This can probably be done with multithreading, which will use less memory and take less time.
- Another way to speed things up is to use [numba](https://numba.pydata.org/) on `stat_fn`'s that are not just simple compositions of numpy functions.

Tagging some folks for feedback on this proof of concept.

### DIFF
--- a/src/mozanalysis/frequentist_stats/bootstrap.py
+++ b/src/mozanalysis/frequentist_stats/bootstrap.py
@@ -3,9 +3,14 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import numpy as np
 import pandas as pd
+import multiprocessing as mp
+import logging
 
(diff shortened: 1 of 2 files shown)


scholtzan:
how does https://github.com/mozilla/mozanalysis/pull/177 compare to this?

mikewilli:
@rebecca-burwei Should this be closed?

PR #188, opened 2023-08-29
fix #187: add sample_size param to enrollments query functions

### LINKED ISSUE #187: Supporting dynamic downsampling for enrollments query
Downsampling is an useful tool for enrollments queries on large experiments, but the easiest way to do this right now involves overriding the entire enrollments query to add a `WHERE sample_id <= {value}` clause without changing the rest of the [template](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/experiment.py#L626-L645).

We should look into adding another parameter to the `_build_enrollments_query_normandy` function so that we can more easily specify a sample rate for enrollment queries. This should allow us to add support in metric-hub and metric-config-parser for a `sample_rate` (or similar) parameter that could get injected into the query from jetstream.

mikewilli:
Adds a `sample_size` parameter to the functions building enrollments queries. This will allow us to update metric-hub configs to support easier downsampling without having to redefine the entire enrollments query. From the config, jetstream can retrieve this sample size and pass it into mozanalysis when building the enrollments query.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -409,6 +409,7 @@ class Experiment:
         custom_exposure_query: Optional[str] = None,
         exposure_signal=None,
         segment_list=None,
+        sample_size: int = None,
     ) -> str:
         """Return a SQL query for querying enrollment and exposure data.
 
@@ -435,11 +436,14 @@ class Experiment:
             segment_list (list of mozanalysis.segment.Segment or str): The user
                 segments to study.
 
+            sample_size (int): Optional integer percentage of clients, used for
+                downsampling enrollments.
+
         Returns:
             A string containing a BigQuery SQL expression.
         """
         enrollments_query = custom_enrollments_query or self._build_enrollments_query(
-            time_limits, enrollments_query_type
+            time_limits, enrollments_query_type, sample_size
         )
 
         if exposure_signal:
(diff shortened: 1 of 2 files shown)


src/mozanalysis/experiment.py
```diff
@@ -637,14 +650,18 @@ def _build_enrollments_query_normandy(self, time_limits: TimeLimits) -> str:
             AND e.submission_date
                 BETWEEN '{first_enrollment_date}' AND '{last_enrollment_date}'
             AND e.event_string_value = '{experiment_slug}'
+            AND e.sample_id <= {sample_size}
         GROUP BY e.client_id, branch
             """.format(
             experiment_slug=self.experiment_slug,
             first_enrollment_date=time_limits.first_enrollment_date,
             last_enrollment_date=time_limits.last_enrollment_date,
+            sample_size=sample_size or 100,
```
danielkberry, line 659: Why the "default none" and "sample_size or 100" pattern instead of setting a default value for the parameter in the function definition?

mikewilli: I actually had it the other way, but noticed that this sort of pattern was used in one of the other enrollment functions. I'm OK with setting the default via function signature though, I'll change it.

danielkberry (approved):
LGTM. What would an end-to-end test of this look like for jetstream config files? Like, how do we test that when someone sets this parameter in a custom config that the expected enrollments SQL query is generated

src/mozanalysis/experiment.py
danielkberry: I don't know if I have an opinion, I was just curious

mikewilli:
>What would an end-to-end test of this look like for jetstream config files? Like, how do we test that when someone sets this parameter in a custom config that the expected enrollments SQL query is generated

Good question. I don't see why we couldn't test for the `sample_id <= {value}` inside jetstream as it contains mozanalysis as a dependency, with a mock for the object that would've been parsed from the config. Then we can also have separate tests for parsing it from the config.

PR #192, opened 2023-09-26
fix glean enrollments query with sample_id

mikewilli:
Glean enrollments query uses `UNNEST(events.events)` as `e` instead of just `events` so we need to change the WHERE clause to remove the `e.` from `sample_id`.


danielkberry (approved):
Makes sense. `sample_id` is a property of the ping, not the event.

PR #193, opened 2023-10-02
handle enrollments and exposures queries for cirrus

mikewilli:
- builds enrollment and exposure queries for Cirrus
- adds an enum for the allowed query types
- adds a test to make sure cirrus queries are built properly

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -22,7 +22,7 @@ class EnrollmentsQueryType(str, Enum):
     CIRRUS = "cirrus"
     FENIX_FALLBACK = "fenix-fallback"
     NORMANDY = "normandy"
-    GLEAN_EVENTS = "glean-events"
+    GLEAN_EVENT = "glean-event"
 
 
 @attr.s(frozen=True, slots=True)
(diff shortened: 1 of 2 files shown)


src/mozanalysis/experiment.py
```diff
@@ -593,43 +602,62 @@ def _build_analysis_windows_query(analysis_windows) -> str:
...
     def _build_exposure_query(
-        self, time_limits: TimeLimits, exposure_query_type: str
+        self, time_limits: TimeLimits, exposure_query_type: EnrollmentsQueryType
     ) -> str:
         """Return SQL to query a list of exposures and their branches"""
-        if exposure_query_type == "normandy":
+        if exposure_query_type == EnrollmentsQueryType.NORMANDY:
             return self._build_exposure_query_normandy(time_limits)
-        elif exposure_query_type == "glean-event":
+        elif exposure_query_type == EnrollmentsQueryType.GLEAN_EVENTS:
```
scholtzan, line 640: I guess this should be `GLEAN_EVENT` and `glean-event` as the string value

mikewilli: Oh good catch, I'll update that.

scholtzan (approved):

PR #195, opened 2023-10-19
Fixes memory leak in `frequentist_stats._resample_and_agg_once`

danielkberry:


mikewilli (approved):
I'm surprised the garbage collector doesn't handle this automatically, but I also don't see any harm here.

PR #196, opened 2023-10-31
Faster decile inference

danielkberry:
This PR provides a new method for quantile inference using a "resampling-free bootstrap". It [provides](https://colab.research.google.com/drive/1WVtx7w9zWSYx5MQTZxQo4gWz73-KlWY0#scrollTo=CWfq0kZYuu5d) >100x speedups over the current implementation (comparison tests in the linked notebook). 

This is valuable as, in some cases, the vast majority of Jetstream analysis duration is spent in the decile statistic.


scholtzan (approved):
lgtm

PR #197, opened 2023-11-14
Changes to allow pre-enrollment analyses

danielkberry:


src/mozanalysis/experiment.py
```diff
@@ -1199,12 +1197,6 @@ def for_ts(
     @first_enrollment_date.validator
     def _validate_first_enrollment_date(self, attribute, value):
         assert self.first_enrollment_date <= self.last_enrollment_date
-        assert self.first_enrollment_date <= self.first_date_data_required
-        assert self.first_enrollment_date <= self.last_date_data_required
-
-    @last_enrollment_date.validator
-    def _validate_last_enrollment_date(self, attribute, value):
-        assert self.last_enrollment_date <= self.last_date_data_required
```
mikewilli, line 1207: It seems like these should still be applicable, no?

src/mozanalysis/experiment.py
```diff
@@ -1227,23 +1219,26 @@ def _validate_last_date_data_required(self, attribute, value):
...
+    For example, ``AnalysisWindow(0, 6)`` is the first week after enrollment
+    and `AnalysisWindow(-8,-1)` is the week before enrollument
 
     Args:
-        start (int): First day of the analysis window, in days since
-            enrollment.
+        start (int): First day of the analysis window, in days relative
+            enrollment. 0 indicates the date of enrollment. Positive numbers
+            are after enrollment, negative are before.
         end (int): Final day of the analysis window, in days since
```
mikewilli, line 1231:
```suggestion
        end (int): Final day of the analysis window, in days relative to enrollment end.
```

src/mozanalysis/experiment.py
```diff
@@ -1227,23 +1219,26 @@ def _validate_last_date_data_required(self, attribute, value):
...
     start = attr.ib(type=int)
     end = attr.ib(type=int)
 
     @start.validator
     def _validate_start(self, attribute, value):
-        assert value >= 0
+        assert (value >= 0) or (value < 0 and self.end < 0)
 
     @end.validator
     def _validate_end(self, attribute, value):
```
mikewilli, line 1244: Should this also be the same sign as `start`?

src/mozanalysis/experiment.py
mikewilli: I'd also add something about how `end` must be the same sign as `start` (or 0). Unless that's not accurate, then don't say it :D

mikewilli (changes requested):
I think a small change to the validation of `end` and a few suggestions for comments, but overall looks good.

src/mozanalysis/experiment.py
danielkberry: With pre-enrollment analyses, the `last_date_data_required` is the day before the `first_enrollment_date`

src/mozanalysis/experiment.py
danielkberry: Good catch, I added a check for that.

src/mozanalysis/experiment.py
danielkberry: Added!

src/mozanalysis/experiment.py
```diff
@@ -1227,27 +1219,34 @@ def _validate_last_date_data_required(self, attribute, value):
 class AnalysisWindow:
     """Represents the range of days in which to measure a metric.
 
-    The range is measured in "days after enrollment", and is inclusive.
+    The range is measured in "days relative enrollment", and is inclusive.
 
-    For example, ``AnalysisWindow(0, 6)`` is the first week after enrollment.
+    For example, ``AnalysisWindow(0, 6)`` is the first week after enrollment
+    and `AnalysisWindow(-8,-1)` is the week before enrollument
```
scholtzan, line 1225:
```suggestion
    and `AnalysisWindow(-8,-1)` is the week before enrollment
```

scholtzan (approved):

PR #198, opened 2023-11-15
Adds p_values for absolute and relative comparisons

danielkberry:
This PR adds p-values to the output of `mozanalysis.bayesian_stats.compare_branches`. P-values are calculated using the `p-value` - `confidence interval` duality: given a bootstrapped confidence interval (as calculated using the existing methodology), this code calculates the corresponding p-value. A validation notebook is available [here](https://colab.research.google.com/drive/1hHnfCtRSwwmXaou6wh3EIOMCXciecTZM?usp=sharing)


danielkberry:
One open question is around downstream impacts. @mikewilli, what changes need to be made to shared data models to accommodate the new field?

PR #199, opened 2023-11-16
Add loading for metrics & data sources defined in outcomes

dzeber:
This will be useful when we want to use metrics or data sources defined in metric-hub outcomes in sizing or other ad-hoc calculations. The current workflow is to copy-paste the metric definition. `ConfigLoader` is now the main way that metric-hub configs are accessed by data scientists working in notebooks.

Other updates:
- Add tests for config.py
  * Currently this relies on pulling specific configs from metric-hub, which could change in the future. The "proper" way would be to mock a version of metric-hub in this repo, but I wasn't sure that was necessary at this point.
- Add API docs for config.py
- Minor cleanup of markup in guide doc page
- Fix setting description for metrics

### DIFF
--- a/docs/api/config.rst
+++ b/docs/api/config.rst
@@ -1,0 +1,9 @@
+:mod:`mozanalysis.config`
+-----------------------------
+
+.. automodule:: mozanalysis.config
+   :members:
+   :private-members: _ConfigLoader
+
+   .. autodata:: ConfigLoader
+
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -51,19 +51,19 @@ We start by instantiating our :class:`mozanalysis.experiment.Experiment` object:
 ``start_date`` is the ``submission_date`` of the first enrollment (``submission_date`` is in UTC). If you intended to study one week's worth of enrollments, then set ``num_dates_enrollment=8``: Normandy experiments typically go live in the evening UTC-time, so 8 days of data is a better approximation than 7.
 
 
-We now gather a list of who was enrolled in what branch and when, and try to quantify what happened to each client. In many cases, the metrics in which you're interested will already be in a metrics library, [Metric Hub](https://github.com/mozilla/metric-hub). If not, then you can define your own - see :meth:`mozanalysis.metrics.Metric` for examples - and ideally submit a PR to add them to Metric Hub for the next experiment. To load a Metric from Metric Hub, for example:
+We now gather a list of who was enrolled in what branch and when, and try to quantify what happened to each client. In many cases, the metrics in which you're interested will already be in a metrics library, `metric-hub <https://github.com/mozilla/metric-hub>`_. If not, then you can define your own---see :class:`mozanalysis.metrics.Metric` for examples---and ideally submit a PR to add them to metric-hub for the next experiment. To load a Metric from metric-hub, for example::
 
     from mozanalysis.config import ConfigLoader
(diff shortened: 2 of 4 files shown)


danielkberry:
So I talked it over with @mikewilli and we think this is a situation where mozanalysis should mostly use metric-config-parser to resolve the metric (i.d., the metric-config-parser should be able to provide an answer to "give me the definition of this metric from this config", instead of mozanalysis having to assemble a definition itself). 

Today, we're pretty close, we can just do something like (error checking omitted for brevity): 

```python
@dataclass
class EmptyConf:
    app_name: str

# ConfigLoader.get_outcome_metric    
def get_outcome_metric(self, metric_slug: str, outcome_slug: str, app_name: str):
    outcome_spec = self.jetstream_configs.spec_for_outcome(
        slug=outcome_slug, platform=app_name
    )
    
    metric_definition = outcome_spec.metrics.get(metric_slug)
    conf = EmptyConf(app_name)
    summaries = metric_definition.resolve(outcome_spec, conf, configs)
    return summaries[0].metric
```

where the `MetricReference.resolve` [method](https://github.com/mozilla/metric-config-parser/blob/main/metric_config_parser/metric.py#L155-L288) does most of the heavy lifting. This is the same function that is called to resolve a metric definition in production jobs (for Jetstream & OpMon) so I think that should be the source of truth for the logic. 

The only thing that's left to do would be to handle the `conf` parameter. Here I've mocked a conf-like object with just the parameter needed, but that's not the proper solution. Interested to get Mike's thoughts here.

dzeber:
Sounds good to me, let me know how you'd like to proceed.

mikewilli:
I don't think @danielkberry's `EmptyConf` is too far off. As noted, `resolve` only really cares about the `app_name`, so I wonder if the solution is to refactor the resolve function to just take `app_name` directly?

Another option, if you want all the metrics from an outcome, might be to add a `resolve` function to `OutcomeSpec`, though this would be more involved. Probably not worthwhile just for this when we're so close already with what already exists.

dzeber:
@danielkberry @mikewilli (Sorry for the delay, I had to backburner this work the past week.) What do you think is the best way forward - would you want to make updates to metric-config-parser before landing this, or else we can use Daniel's approach and update later?

> Another option, if you want all the metrics from an outcome

I agree this is probably not worthwhile at this point, as the use case I have in mind is to get specific metrics for sizing.

mikewilli:
@dzeber In the interest of starting with the path of least resistance, why don't we proceed with @danielkberry's `EmptyConf` solution (maybe another name would be better.. `AppOnlyConfiguration`, `MinimalConfiguration`,  ..? open to ideas, but also it's not a huge issue).

This will allow you to use the existing `resolve` to retrieve and use definitions for outcome metrics without requiring any external updates (i.e., to metric-config-parser). And I am concerned that updating the `resolve` function signature to only take `app_name` would require a bunch of downstream changes for consumers of metric-config-parser, so if we're going to make changes there I'd rather do so after more consideration.

Let me know if you have any trouble integrating the code @danielkberry shared into this PR, happy to help.

dzeber:
Quick update: I won't have time to finish this up before holidays but I'll pick it up when I get back in Jan.

dzeber:
I've updated the outcome metric loading using @danielkberry 's approach.

I also fixed a typo where description wasn't getting set in `get_metric`.

mikewilli (approved):
Looks good to me! Doc updates and tests too!

I say we ship it, and we can always iterate on the approach as needed. :shipit:

PR #201, opened 2023-12-05
Fetching outcomes from generic metric source

scholtzan:
A few ideas of how we could get outcome data more generically. I guess currently `jetststream` is the only place where outcome data is defined, but it is not impossible that we have other applications/sources where outcome data is defined. We can have a more generic way to specify from which repo/path we want to load data from.

### DIFF
--- a/src/mozanalysis/config.py
+++ b/src/mozanalysis/config.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from typing import Optional
+from typing import List, Optional
 
 from jinja2 import UndefinedError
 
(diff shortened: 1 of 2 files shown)


src/mozanalysis/config.py
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from typing import Optional
+from typing import List, Optional
```
mikewilli, line 5: I was going to suggest we use generics for type hints instead of importing `List`, but apparently `mozanalysis` is on python 3.8 so maybe that's the more pressing issue...

mikewilli (approved):
One question: does this work for metrics whose data source is defined elsewhere, like `main`? If not, we'd still need to be able to call `resolve` to get that definition.

Either way this is great!

dzeber (approved):
Thanks for this! This looks great and is definitely a more extensible approach

PR #203, opened 2024-01-27
Expand empirical effect size results

dzeber:
This adds metric mean value and relative effect size to the empirical sizing output dict.

Additionally:
- Simplify the code to build the result dict
- Add a test for empirical sizing


m-d-bowerman (approved):
lgtm

PR #206, opened 2024-02-23
Improve readability of sizing functions

### LINKED ISSUE #204: Improve readability of sizing functions
Currently, the functions in `sample_size.py` ([link](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/frequentist_stats/sample_size.py)) return either dictionaries, lists of DataFrames, or lists of dictionaries. We should create a class to hold sample size calculation results to make handling the results more user-friendly.

jaredsnyder:
This PR is adding wrapper classes for results returned from the objects in sample_size_calc.py.  The idea is to add functionality to make it easier to work with results while maintaining backward compatibility

Fixes #204

### DIFF
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -151,7 +151,7 @@ class EmpericalEffectSizeResultsHolder(ResultsHolder):
             .format("{:.2%}", subset="rel_effect_size")
         )
 
-    def get_dataframe(self, style: bool = None) -> Union[pd.DataFrame, Styler]:
+    def get_dataframe(self, style: bool = True) -> Union[pd.DataFrame, Styler]:
         """returns dataframe for results from empirical_effect_size_sample_size_calc
 
         Arguments:
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -26,7 +26,7 @@ from mozanalysis.utils import get_time_intervals
 class SampleSizeResultsHolder(UserDict):
     """
     Object to hold results from different methods.  It extends
-    the dictionary objects so that users can interact with it 
+    the dictionary objects so that users can interact with it
     with a dictionary with the same keys/values as before, making
     it backward compatible.  The dictionary functionality is extended to include
     additional attributes to hold metadata, a method for plotting results and
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -354,10 +354,12 @@ def sample_size_curves(
         **solver_kwargs (dict): Arguments necessary for the provided solver.
 
     Returns:
-        A dictionary of pd.DataFrame objects. An item in the dictionary is
+        SampleSizeCurveResultHolder: The data attribute contains a dictionary
+        of pd.DataFrame objects. An item in the dictionary is
         created for each metric in metric_list, containing a DataFrame of sample
         size per branch, number of clients that satisfied targeting, and population
         proportion per branch at each value of the iterable parameter.
+        Additional methods for ease of use are documented in the class.
     """
     params = {"effect_size": effect_size, "power": power, "alpha": alpha}
     sim_var = [k for k, v in params.items() if type(v) in [list, np.ndarray, pd.Series]]
@@ -415,9 +417,11 @@ def difference_of_proportions_sample_size_calc(
             each columns.
 
     Returns:
-        A dictionary. Keys in the dictionary are the metrics column names from
+        SampleSizeResultsHolder: The data attribute contains a dictionary.
+        Keys in the dictionary are the metrics column names from
         the DataFrame; values are the required sample size per branch to achieve
-        the desired power for that metric.
+        the desired power for that metric. Additional methods for ease of use
+        are documented in the class.
     """
 
     def _get_sample_size_col(col):
@@ -478,9 +482,11 @@ def z_or_t_ind_sample_size_calc(
             each columns.
 
     Returns:
-        A dictionary. Keys in the dictionary are the metrics column names from
+        SampleSizeResultsHolder: The data attribute contains a dictionary.
+        Keys in the dictionary are the metrics column names from
         the DataFrame; values are the required sample size per branch to achieve
-        the desired power for that metric.
+        the desired power for that metric. Additional methods for ease of use
+        are documented in the class.
     """
     tests = {
         "normal": zt_ind_solve_power,
@@ -561,10 +567,12 @@ def empirical_effect_size_sample_size_calc(
             distribution of effect sizes observed in historical data.
 
     Returns:
-        A dictionary. Keys in the dictionary are the metrics column names from
+        EmpericalEffectSizeResultsHolder: The data attribute contains a dictionary.
+        Keys in the dictionary are the metrics column names from
         the DataFrame; values are dictionaries containing the required sample size
         per branch to achieve the desired power for that metric, along with
-        additional information.
+        additional information. Additional methods for ease of use
+        are documented in the class.
     """
 
     def _mann_whitney_solve_sample_size_approximation(
@@ -690,9 +698,11 @@ def poisson_diff_solve_sample_size(
             each columns.
 
     Returns:
-        A dictionary. Keys in the dictionary are the metrics column names from
+        SampleSizeResultsHolder: The data attribute contains a dictionary.
+        Keys in the dictionary are the metrics column names from
         the DataFrame; values are the required sample size per branch to achieve
-        the desired power for that metric.
+        the desired power for that metric.  Additional methods for ease of use
+        are documented in the class
     """
 
     def _get_sample_size_col(col):
--- a/src/mozanalysis/frequentist_stats/sample_size.py
+++ b/src/mozanalysis/frequentist_stats/sample_size.py
@@ -3,17 +3,17 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 """module for sample size calculations"""
 
-from datetime import datetime
 from collections import UserDict
+from datetime import datetime
 from math import pi
 
-from scipy.stats import norm
-from statsmodels.stats.power import tt_ind_solve_power, zt_ind_solve_power
-from statsmodels.stats.proportion import samplesize_proportions_2indep_onetail
+import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 from pandas.io.formats.style import Styler
-import matplotlib.pyplot as plt
+from scipy.stats import norm
+from statsmodels.stats.power import tt_ind_solve_power, zt_ind_solve_power
+from statsmodels.stats.proportion import samplesize_proportions_2indep_onetail
 
 from mozanalysis.bq import BigQueryContext
 from mozanalysis.experiment import TimeSeriesResult
@@ -31,7 +31,9 @@ class ResultsHolder(UserDict):
     it backward compatible.
     """
 
-    def __init__(self, *args, metrics: dict = None, params: dict = None, **kwargs):
+    def __init__(
+        self, *args, metrics: dict | None = None, params: dict | None = None, **kwargs
+    ):
         """
         Args:
             metrics (list, optional): List of metrics used to generate the results.
@@ -255,9 +257,9 @@ class SampleSizeCurveResultHolder(ResultsHolder):
         self,
         input_data: pd.DataFrame = None,
         show_population_pct: bool = True,
-        simulated_values: List[float] = None,
+        simulated_values: list[float] | None = None,
         append_stats: bool = False,
-        highlight_lessthan: List[float] = None,
+        highlight_lessthan: list[float] | None = None,
         trim_highlight_threshold: float = 0.15,
     ) -> Styler:
         """
@@ -347,7 +349,9 @@ class SampleSizeCurveResultHolder(ResultsHolder):
             # Only 3 cutoffs suported. Any others are silently dropped
             highlight_lessthan = sorted(highlight_lessthan)[:3]
             # Apply from highest to lowest cutoff so that lower cutoffs overwrite higher
-            for lim, colour in list(zip(highlight_lessthan, highlight_colours))[::-1]:
+            for lim, colour in list(
+                zip(highlight_lessthan, highlight_colours, strict=False)
+            )[::-1]:
                 disp = disp.highlight_between(
                     subset=simulated_values,
                     right=lim,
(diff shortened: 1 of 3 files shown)


jaredsnyder:
For the simplest methods that just return a dict of dicts I created `SampleSizeResultsHolder`.  It is just an extension of the dict class that has some extra methods and attributes, but is set up so that if you interact with it like a dict it works exactly the same as it did before.  Added a nice graphing function where the output looks like this:

image

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +22,114 @@
...
+            params (dict, optional): _description_. Defaults to None.
+        """
+        super().__init__(*args, **kwargs)
+        self._metrics = metrics
+        self._params = params
+        # create this attribute to hold only the results data
+        # this dict is what was returned from the sample size
+        # methods historically
+        self.results = {k: v for k, v in self.data.items()
+                        if k not in ['metrics', 'params']}
```
dzeber, line 51: What about keeping this in `data` rather than creating a separate `results` dict, ie. modifying `data`? That way all the dunder methods like `__getitem__` work out of the box without needing to be overridden.

dzeber (commented):
This looks to me like a good approach. We may want to move in a direction to change those underlying return formats eventually, but right now good to remain backward-compatible.

With this pattern, we could also subclass the `SampleSizeResultsHolder` if we want specific additional functionality for a particular sample size tool.

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: Good catch, fix in most recent commit

jaredsnyder:
This is what the `pretty_results` function output looks like
Screenshot 2024-02-26 at 11 31 09 AM

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+        }
+        nice_result = self.make_friendly_name(result_name)
+        df = self.get_dataframe().rename(index=nice_metric_map)
+        df[result_name].plot(kind="bar")
+        plt.ylabel(nice_result)
+        plt.xlabel("Metric")
+        plt.show(block=False)
+
+
+class EmpericalEffectSizeResultsHolder(ResultsHolder):
```
teonbrooks, line 120:
```suggestion
class EmpiricalEffectSizeResultsHolder(ResultsHolder):
```

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -214,7 +606,7 @@ def empirical_effect_size_sample_size_calc(
     alpha: float = 0.05,
     parent_distribution: str = "normal",
     plot_effect_sizes: bool = False,
-) -> dict:
+) -> EmpericalEffectSizeResultsHolder:
```
teonbrooks, line 609:
```suggestion
) -> EmpiricalEffectSizeResultsHolder:
```

tests/frequentist_stats/test_sample_size_calc.py
```diff
@@ -74,3 +75,199 @@ class FakeMetric:
         np.testing.assert_allclose(
             r["population_percent_per_branch"], 0.844, atol=0.001
         )
+
+
+def test_emperical_effect_size_results_holder(fake_ts_result):
```
teonbrooks, line 80:
```suggestion
def test_empirical_effect_size_results_holder(fake_ts_result):
```

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -247,10 +639,12 @@ def empirical_effect_size_sample_size_calc(
             distribution of effect sizes observed in historical data.
 
     Returns:
-        A dictionary. Keys in the dictionary are the metrics column names from
+        EmpericalEffectSizeResultsHolder: The data attribute contains a dictionary.
```
teonbrooks, line 642:
```suggestion
        EmpiricalEffectSizeResultsHolder: The data attribute contains a dictionary.
```

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -336,7 +730,16 @@ def _mann_whitney_solve_sample_size_approximation(
...
 
-    return size_dict  # TODO: add option to return a DataFrame
+    params = {
+        "quantile": quantile,
+        "power": power,
+        "alpha": alpha,
+        "parent_distribution": parent_distribution,
+    }
+
+    return EmpericalEffectSizeResultsHolder(
```
teonbrooks, line 740:
```suggestion
    return EmpiricalEffectSizeResultsHolder(
```

teonbrooks (commented):
these are just some spelling corrections I came across while browsing. this isn't a code reveiw. if you would like an extra set of eyes, i can do so

jaredsnyder:
> these are just some spelling corrections I came across while browsing. this isn't a code reveiw. if you would like an extra set of eyes, i can do so

Whoops, thanks for catching that!  Pushed a commit to fix.

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+            show_population_pct (bool, optional): If true, percentage of population
+                will be used in output rather than raw counts. Defaults to True.
+            simulated_values (list, optional): List of simulated values to subset
+                output to. If None, all values will be included. Defaults to None.
+            append_stats (bool, optional): _description_. If True, summary statistics on
+                the raw metrics are provided. Defaults to False.
+            highlight_lessthan (list[tuple], optional): List of
+                (<cutoff (float)>, <color (string)>) tuples.
+                Sample size entries less than <cutoff> will be highlighted in <color>
+                in the displayed table. Defaults to None.
```
dzeber, line 273: Can be dropped

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+            pretty_df = self.raw_df[subset_col].unstack()
+        else:
+            pretty_df = self.raw_df[subset_col].unstack()[simulated_values]
+
+        # get stats if input_data is non-null and append_stats is True
+        if append_stats:
+            if input_data is None:
+                raise ValueError("append_stats is true but no raw data was provided")
+            self.set_raw_data_stats(input_data)
+            pretty_df = pd.concat([pretty_df, self._raw_data_stats], axis="columns")
```
dzeber, line 300: Similarly, in keeping with `get_dataframe` for the other result types, maybe appending stats should happen in `get_styled_dataframe`, but the output of this method just contains the main results returned by the sizing tool

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+        overall_stats = (
+            input_data[[m.name for m in self._metrics]]
+            .agg(
+                [
+                    "mean",
+                    "std",
+                    lambda d: d[d <= d.quantile(outlier_percentile)].mean(),
+                    lambda d: d[d <= d.quantile(outlier_percentile)].std(),
+                ]
+            )
```
dzeber, line 239: These stats are also computed in all the solver functions (but don't get returned). Showing these stats is not specific to the sample size curves tool. I wonder if this should be a separate function at the top level of the module, and then it could get called in the `_get_sample_size_col` functions. We could then return these stats in another attr of the ResultsHolder. That would also help avoid having the trimming logic replicated in multiple places.

Within the context of this PR I think it's fine to leave this here, but we could consider making this change in a future one.

@m-d-bowerman What do think about updating the solver functions in this way? I usually recompute these stats after the fact when sizing because I want to see them alongside the sizing results

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+
+    def __init__(self, *args, metrics: dict = None, params: dict = None, **kwargs):
+        """
+        Args:
+            metrics (dict, optional): _description_. Defaults to None.
+            params (dict, optional): _description_. Defaults to None.
+        """
+        super().__init__(*args, **kwargs)
+        self._metrics = metrics
+        self._params = params
```
dzeber, line 43: Should we add property getters (`@property`) for these?

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+        ).abs() / overall_stats["std"]
+        self._raw_data_stats = overall_stats
+
+    def get_dataframe(
+        self,
+        input_data: pd.DataFrame = None,
+        show_population_pct: bool = True,
+        simulated_values: list = None,
+        append_stats: bool = False,
+    ) -> pd.DataFrame:
```
dzeber, line 257: In keeping with `get_dataframe` for the other result types, I wonder if this should have no args and just return the main results returned by the sizing function. The args to control display can instead by used in `get_styled_dataframe`. Cf comments below

dzeber (commented):
Looking good!

@m-d-bowerman are there any downstream dependencies (eg in Nimbus) we need to be aware of? I think you were working on an auto-sizing tool?

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+        nice_metric_names = [
+            (
+                el.friendly_name
+                if hasattr(el, "friendly_name")
+                else self.make_friendly_name(el.name)
+            )
+            for el in self._metrics
+        ]
+        nice_metric_map = {
+            el.name: friendly_name
```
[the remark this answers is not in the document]
jaredsnyder: Yeah that's a good change

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+
+    def get_dataframe(self) -> pd.DataFrame:
+        """returns dataframe for results from empirical_effect_size_sample_size_calc
+
+        Arguments:
+            style (bool): If true, return a cleaned up and formatted pandas Styler.
+            Otherwise return a dataframe
+        Returns:
+            pd.DataFrame or Styler: dataframe containing results if style is False
+            If true it returns a Styler object
```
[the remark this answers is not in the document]
jaredsnyder: oops, good catch

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: I think that'd be a good change.  Maybe we should go one step further and make a `dataframe` property.  I'll work on it

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: This is a good change but for fear of messing up metric functions like `z_or_t_ind_sample_size_calc` I'd prefer to leave it for another PR.  If you or Michael feel strongly otherwise we can talk about it.

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+                        provided
+
+        Returns:
+            pd.DataFrame : dataframe consolidating data
+        """
+        # choose which column to output
+        if show_population_pct:
+            subset_col = "population_percent_per_branch"
+        else:
+            subset_col = "sample_size_per_branch"
```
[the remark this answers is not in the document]
jaredsnyder: fixed

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -12,13 +23,374 @@
...
+    def __init__(self, *args, **kwargs):
+        """
+        Args:
+            metrics (dict, optional): _description_. Defaults to None.
+            params (dict, optional): _description_. Defaults to None.
+        """
+        super().__init__(*args, **kwargs)
+        # need to modify results so it matches the old format
+        self.raw_df = pd.concat(
+            [el.set_index("effect_size", append=True) for el in self.data.values()]
```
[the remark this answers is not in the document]
dzeber: Right, but should it be `sim_var` instead?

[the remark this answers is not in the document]
jaredsnyder: Ah good catch, fixed

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -20,6 +23,344 @@
...
+        }
+
+    @property
+    def metrics(self):
+        """the metrics property"""
+        return self._metrics
+
+    @property
+    def params(self):
+        """the self property"""
```
dzeber, line 61:
```suggestion
        """the params property"""
```

dzeber: or maybe copy the descriptions from the __init__ docstrings into these ones?

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -20,6 +23,344 @@
...
+        return self.style_empirical_sizing_result(df)
+
+
+class SampleSizeCurveResultHolder(ResultsHolder):
+    def __init__(self, *args, **kwargs):
+        """
+        Args:
+            metrics (dict, optional): _description_. Defaults to None.
+            params (dict, optional): _description_. Defaults to None.
+        """
```
dzeber, line 215: remove?

src/mozanalysis/frequentist_stats/sample_size.py
```diff
@@ -20,6 +23,344 @@
...
+
+        # get stats if input_data is non-null and append_stats is True
+        if append_stats:
+            if input_data is None:
+                raise ValueError("append_stats is true but no raw data was provided")
+            self.set_raw_data_stats(input_data)
+            pretty_df = pd.concat([pretty_df, self._raw_data_stats], axis="columns")
+
+        if simulated_values is None:
+            simulated_values = self._params["simulated_values"]
```
dzeber, line 316: Is this still needed?

dzeber (approved):
I had a few minor comments, but overall LGTM!

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: This is a really funny mistake, I like the idea of using the descriptions from the init docstring

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: Yeah, replaced with a comment highlighting the use of the ResultsHolder __init__

src/mozanalysis/frequentist_stats/sample_size.py
jaredsnyder: Nope, good catch!

PR #208, opened 2024-02-27
add deprecation warning for metric helper imports

### LINKED ISSUE #171: Deprecate overloaded import of metric-hub metrics
It's confusing for DS who are conducting sizing exercises to know when to import metrics from metric-hub and when to import [from mozanalysis](https://github.com/mozilla/mozanalysis/blob/main/src/mozanalysis/metrics/desktop.py). I think we should move to a workflow where DS import directly from metric-hub and remove this re-importing.

jaredsnyder:
This PR fixes #171 .  Deprecation warnings were added to all the helper variables in metrics.  When one tries to import, the output looks like this:

```
>>> from mozanalysis.metrics.desktop import clients_daily
/Users/jsnyder/code/mozanalysis/src/mozanalysis/metrics/desktop.py:13: DeprecationWarning: 
    metrics and data sources created in mozanalysis are deprecated
    please create directly from metric-hub with ConfigLoader like

    from mozanalysis.config import ConfigLoader
    metric=ConfigLoader.get_metric(metric_slug="active_hours",app_name="firefox_desktop")

    and data sources like
    data_source=ConfigLoader.get_data_source(data_source_slug="active_hours",
                                                app_name="firefox_desktop")

    
  warnings.warn(
  ```

### DIFF
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -7,24 +7,7 @@
 # in https://github.com/mozilla/metric-hub
 
 from mozanalysis.config import ConfigLoader
-import warnings
 
-warnings.simplefilter("default")
-warnings.warn(
-    """
-    metrics and data sources created in mozanalysis are deprecated
-    please create directly from metric-hub with ConfigLoader like
-
-    from mozanalysis.config import ConfigLoader
-    metric=ConfigLoader.get_metric(metric_slug="active_hours",app_name="firefox_desktop")
-
-    and data sources like
-    data_source=ConfigLoader.get_data_source(data_source_slug="active_hours",
-                                                app_name="firefox_desktop")
-
-    """,
-    DeprecationWarning,
-)
 
 clients_daily = ConfigLoader.get_data_source("clients_daily", "firefox_desktop")
 
--- a/src/mozanalysis/metrics/desktop.py
+++ b/src/mozanalysis/metrics/desktop.py
@@ -7,6 +7,24 @@
 # in https://github.com/mozilla/metric-hub
 
 from mozanalysis.config import ConfigLoader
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/mozanalysis/metrics/desktop.py [hunk 1]
```diff
@@ -7,6 +7,24 @@
 # in https://github.com/mozilla/metric-hub
 
 from mozanalysis.config import ConfigLoader
+import warnings
+
+warnings.simplefilter("default")
+warnings.warn(
+    """
+    metrics and data sources created in mozanalysis are deprecated
```
danielkberry, line 15: If we're adding the same warning to a bunch of different files, should we abstract that string out to a single place and just import into each file? something like:

```python
from mozanalysis.warning_messages import mozanalysis_metric_deprecation

warnings.simplefilter("default")
warnings.warn(mozanalysis_metric_deprecation)
```

in each file and then the actual warning string defined in `warning_messages` or something

jaredsnyder: Good idea, instead of importing the warning message I just put it in the `__init__.py` so that if you import any of the metric files it raises the warning.

jaredsnyder:
@danielkberry Should we also add a deprecation warning to `segements` ?

m-d-bowerman:
> @danielkberry Should we also add a deprecation warning to `segements` ?

Yeah, a deprecation warning should be added for `Segment` and `SegmentDataSource`, too

jaredsnyder:
> > @danielkberry Should we also add a deprecation warning to `segements` ?
> 
> Yeah, a deprecation warning should be added for `Segment` and `SegmentDataSource`, too

Just pushed this

danielkberry (approved):
LGTM

Issue #11, opened 2018-11-06
Add version info


robhudson:
When reports are created it would be helpful if mozanalysis (and other libs) were able to show their version info in the report for reproducability.

Something like:
```
>>> import mozanalysis
>>> mozanalysis.VERSION
2018.11.1
```

felixlawrence:
@tdsmith: we could probably hook in https://pypi.org/project/setuptools-scm/ to expose this

PR #209, opened 2024-02-27
feat: add __version__ attribute

jaredsnyder:
This is meant to address this issue: https://github.com/mozilla/mozanalysis/issues/11

One can access the version using the __version__ attribute:

```
>>> import mozanalysis
>>> mozanalysis.__version__
'2023.11.4.dev12+g230bf07.d20240227'
```

It looks like the version is generated by `setuptools_scm`, and it gets grabbed by `importlib.metadata`.  We could add a little logic just to grab the date part of it so it matches with the git tags.

### DIFF
--- a/src/mozanalysis/__init__.py
+++ b/src/mozanalysis/__init__.py
@@ -7,6 +7,12 @@
 # We use it to expose modules available in mozanalysis package.
 # https://github.com/mozilla/jetstream
 import os
+from importlib.metadata import version, PackageNotFoundError
+
+try:
+    __version__ = version("mozanalysis")
+except PackageNotFoundError:
+    __version__ = "unknown version"
 
 __all__ = []
 basedir = os.path.dirname(__file__)


jaredsnyder:
The tests are failing because of an issue with python 3.8 and the latest version of pytest.  I can fix the issue by setting the pytest version to be <7 but I don't know if that makes sense for such a trivial change

dzeber:
> The tests are failing because of an issue with python 3.8 and the latest version of pytest. I can fix the issue by setting the pytest version to be <7 but I don't know if that makes sense for such a trivial change

I think the long-term solution is probably to bump the python version for testing. I opened #210 for this.

dzeber (approved):
LGTM. Tagging @mikewilli to double-check if this would have any downstream impacts

mikewilli:
Unless this is urgent, we can probably block it on #210? If this is urgent or #210 turns out to be more annoying than expected, go ahead and pin pytest, but as a separate PR that can be merged first.

mikewilli:
I should clarify that I don't see any downstream impacts, adding `__version__` should be harmless.

jaredsnyder:
> Unless this is urgent, we can probably block it on #210? If this is urgent or #210 turns out to be more annoying than expected, go ahead and pin pytest, but as a separate PR that can be merged first.

Not urgent, I agree we should wait for #210

Issue #210, opened 2024-02-27
Update Python version under which tests are run


dzeber:
Currently [tests are run](https://github.com/mozilla/mozanalysis/blob/main/tox.ini) under Python 3.8. This sometimes causes issues as dependencies aren't pinned and no Python verison is specified in setup.py.

We should update the tests to default to a more recent Python version.

jaredsnyder:
Does tox make it easy to test multiple versions?  It'd be nice to specify which versions are supposed in the `setup.py` and then test all of them (at least at the minor version level) via tox

mikewilli:
Yes we should definitely specify a python version. I'd suggest 3.10 as that's what Jetstream uses and mozanalysis seems to work ok in there. I also think 3.10 is the most common version I've seen on projects I've touched lately. 3.11 would be fine too, just would want to do extra testing.

And while I think tox can handle multiple versions pretty easily (just specify in the `envlist` in `tox.ini`), I think we may want to settle on a single version to simplify CI (which uses docker containers pinned to specific python versions).

dzeber:
I was thinking 3.10 as well. I agree we should have a main supported version which is used in CI - we can always run tests locally for other Python versions by passing a version arg to tox.

jaredsnyder:
Been taking a stab this morning and I think `pytest-black` might be broken by pytest version 8.  There is a breaking change with hooks in pytest v8 and `pytest-black` hasn't been updated since 2020.  Do we want to explore other options for enforcing black formatting?  I could look for another library to replace `pytest-black` or we could use pre-commit?

jaredsnyder:
Well it seems like there isn't a limit on the pytest version here, so maybe something is wrong on my end... https://docs.pytest.org/en/7.1.x/reference/plugin_list.html

jaredsnyder:
In setup.py, python_requires set to allow 3.10 and 3.11

danielkberry:
We should probably aim to 1) keep our analysis tools in sync and 2) limit the set of supported versions to reduce the testing load. I'd be ok with formally supporting 3.10 and 3.11 now.

PR #211, opened 2024-02-28
Set python version, Replace black and flake8 with RUFF

jaredsnyder:
pytest-black appears to be dead, and a recent update to pytest has caused it to start throwing error.  This PR replaces it with pytest-ruff, and because ruff is both a linter and a formatter it replaces black and flake8. 

Meant to address #210

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,6 @@
 test_settings: &test_settings
   steps:
     - checkout
-    - run: &install_packages
-        name: Install system packages
-        command: |
-          # pyspark needs to run Java, so we install openjdk.
-          apt-get update
-          apt-get install -y default-jre-headless
     - run: &run_tox_environment_matching_circleci_job_name
         name: Run tox job
         command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,17 @@ test_settings: &test_settings
         command: |
           pip install tox
           tox -e $CIRCLE_JOB
-    - codecov/upload
+    - run:
+        name: Submit code coverage data
+        command: |
+          curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
+          curl -Os https://uploader.codecov.io/latest/linux/codecov
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
+          gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+          shasum -a 256 -c codecov.SHA256SUM
+          chmod +x codecov
+          ./codecov -F "$(basename $PWD | sed s/[^a-z]/_/g)"
 
 
 
@@ -22,10 +32,9 @@ test_settings: &test_settings
 # Jobs: see https://circleci.com/docs/2.0/jobs-steps/
 ####################
 
-version: '2.1'
-orbs:
-  codecov: codecov/codecov@4.0.1
+version: 2
 jobs:
+
   py310:
     <<: *test_settings
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,8 +52,8 @@ jobs:
     - run:
         name: Run linting
         command: |
-          pip install flake8
-          flake8 src/ tests/
+          pip install ruff
+          ruff check src/ tests/
 
   # Runs when the repository is tagged for release; see the workflows section
   # below for trigger logic.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,11 +44,11 @@ jobs:
   py310:
     <<: *test_settings
     docker:
-      - image: python:3.10-bookworm
+      - image: cimg/python:3.10
 
   lint:
     docker:
-      - image: python:3.10-bookworm
+      - image: cimg/python:3.10
     steps:
     - checkout
     - run:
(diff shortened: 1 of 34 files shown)


.circleci/config.yml
```diff
@@ -39,27 +39,27 @@ test_settings: &test_settings
 version: 2
 jobs:
 
-  py38:
+  py310:
     <<: *test_settings
     docker:
-      - image: python:3.8-buster
+      - image: python:3.10-buster
```
mikewilli, line 45: While we're switching images, I'd suggest switching to bookworm (or at least bullseye) -- buster is scheduled to end support 30 June 2024.

jaredsnyder: Fixed!

jaredsnyder: Actually I am seeing this error with codecov:

```
{'detail': ErrorDetail(string='Missing "owner" argument. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

Any idea what's going on there?

.circleci/config.yml
```diff
@@ -39,27 +39,27 @@ test_settings: &test_settings
 version: 2
 jobs:
 
-  py38:
+  py310:
     <<: *test_settings
     docker:
-      - image: python:3.8-buster
+      - image: python:3.10-bookworm
```
mikewilli, line 45: I know I just suggested you use 3.10-bookworm, but I realized that this is exclusively for CI, so I think it's suggested to use the CircleCI provided convenience images, in this case: https://circleci.com/developer/images/image/cimg/python.
```suggestion
      - image: cimg/python:3.10
```

.circleci/config.yml
```diff
@@ -23,12 +17,14 @@ test_settings: &test_settings
...
-          # Upload to codecov and flag (-F) with the name of the job.
-          bash <(curl -s https://codecov.io/bash) -F $CIRCLE_JOB
+          curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
+          curl -Os https://uploader.codecov.io/latest/linux/codecov
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
+          gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+          shasum -a 256 -c codecov.SHA256SUM
+          chmod +x codecov
+          ./codecov -F "$(basename $PWD | sed s/[^a-z]/_/g)"
```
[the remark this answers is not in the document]
mikewilli: Oh and if you do want to try the orb and are not sure what needs to be done, I'm happy to help.

ruff.toml
```diff
@@ -0,0 +1,5 @@
+line-length=88
+exclude = [".git","__pycache__"]
+
+[lint]
+ignore= ["E741"]
```
mikewilli, line 5: You'll also want to have a `select` array here of rules to include, otherwise I don't think `ruff` will have any rules to check against.

From the docs, a reasonable starting place that isn't too picky:
```
select = [
  # pycodestyle
  "E",
  # Pyflakes
  "F",
  # pyupgrade
  "UP",
  # flake8-bugbear
  "B",
  # flake8-simplify
  "SIM",
  # isort
  "I",
]
```

telemetry-airflow uses 
```
select = [
    "E", # pycodestyle
    "W", # pycodestyle
    "F", # Pyflakes
    "B", # flake8-bugbear
    "C4", # flake8-comprehensions
    "D", # flake8-docstrings
    "I", # isort
    "SIM", # flake8-simplify
    "TCH", # flake8-type-checking
    "TID", # flake8-tidy-imports
    "Q", # flake8-quotes
    "UP", # pyupgrade
    "PT", # flake8-pytest-style
    "RUF", # Ruff-specific rules
]
```

Experimenter uses
```
select = [
    "A",
    "B",
    "C4",
    "DJ",
    "E",
    "F",
    "I",
    "ISC",
    "N",
    "PIE",
    "Q",
    "RET",
    "RSE",
    "RUF",
    "SIM",
    "TID",
    "UP",
    "W",
    "YTT",
    "PTH",
]
```

jaredsnyder: Ah whoops.  Ended up going with the telemetry airflow setting, minus the docstring check because almost none of the test functions have them.  Created #213 to address this because I think it'd be good to enforce docstrings going forward.

jaredsnyder:
The size of this PR exploded because of all the changes Ruff made 😅

ruff.toml
jaredsnyder: The only changes that are worth a closer look are where it replaced `.format` with an f-string.  A lot of them are used in queries and I don't know how well-tested the query statements are.  I looked through it and they all look fine but it could use a second pair of eyes.  Might also make sense to just make it into another PR at this point 🤷 

Also it left ~15 `.format` statements and I'm not sure why, don't know if there is any point to trying to get rid of them here or just let it go

jaredsnyder:
Okay changes reverted and select is set to `['E', 'F']`

ruff.toml
jaredsnyder: Ended up just doing E and F

README.md
```diff
@@ -26,7 +37,7 @@ or the tests you run. Options after `--` or positional arguments are forwarded t
 For example, you can run:
 
 * `tox -e lint` to lint
-* `tox -e py37 -- -k utils` to only run tests with "utils" somewhere in the name, on Python 3.7
+* `tox -e py310 -- -k utils` to only run tests with "utils" somewhere in the name, on Python 3.7
```
mikewilli, line 40:
```suggestion
* `tox -e py310 -- -k utils` to only run tests with "utils" somewhere in the name, on Python 3.10
```

mikewilli (approved):
Saw one missed python 3.7 reference in the README but otherwise LGTM.

 I think we've discussed this enough that any further work can be done separately (like adding/fixing more ruff rules, and maybe having a list of dependencies somewhere in the repo).

Issue #202, opened 2023-12-19
Better dependency resolution


mikewilli:
There was a (replicable) issue in colab recently where `pip install mozanalysis` was getting incompatible dependencies (in this case: `polyfactory==2.13.0` and `typing-extensions==4.5.0`), resulting in an error when trying to use it. In a separate local venv, I got `polyfactory==2.13.0` and `typing-extensions==4.9.0`, which works.

Python is notoriously difficult when it comes to dependency management and package publication, but we should investigate whether there's anything we can do to at least try to address this.

jaredsnyder:
How do folks feel about poetry?  I have used it a lot in the past and liked how clean it made dealing with dependencies.  Here is the pro/con list as I see it:

Pro:
- Very reliable dependency resolution
- We could consolidate several config files into a `pyproject.toml` (though this is also possible with setuptools, but I've found it to be a bit more difficult than with poetry)
- It also manages virtualenvs
- Changes would only affect developers

Cons:
- Dependency resolution is slow, when you add a new package it can take several minutes to figure it out
- Developers would have to modify their workflow

mikewilli:
I'm fine with poetry and we use it on Experimenter, but it's also a different workflow that folks may not be used to. I'd probably lean towards using the setup that Jetstream has -- pip-tools generates a requirements.txt from requirements.in, and dependabot creates dep update PRs -- it may be an equally different workflow to poetry, but at least there'd be consistency with Jetstream and other similar tooling (like metric-config-parser).

Personally I don't have a strong opinion. I think there's an argument for either side so I'd be interested in getting input and/or votes from the regular contributors before making a decision. What do you think about putting together an ADR to list out the pros/cons of the options and make a recommendation? That would give people an opportunity to comment in the form of a PR. If you're open to the idea but unfamiliar with the ADR process I'd be happy to walk you through it (it's pretty straightforward I think).

jaredsnyder:
Yeah I don't think either is obviously better so consistency makes the Jetstream way a better option.

jaredsnyder:
I could put something together!  I'm not familiar with the process so I'll take you up on your offer to help

jaredsnyder:
As per https://github.com/mozilla/mozanalysis/pull/214#discussion_r1524965545, check to make sure there aren't unnecessary pinned dependencies

PR #214, opened 2024-02-29
Formatting changes to enable more Ruff Rules

### LINKED ISSUE #213: Add more checks to Ruff for cleaner code
Currently Ruff is only checking Pyflakes and pycodestyle rules.  Other projects and mozilla check many more rules that could help keep our code cleaner.  See: https://github.com/mozilla/mozanalysis/pull/211#discussion_r1506714575

jaredsnyder:
fixes #213


jaredsnyder:
Ruff does not yet support Sphinx-style docstrings :upside_down_face: https://github.com/astral-sh/ruff/issues/6606

jaredsnyder:
Removed the docstring check because sphinx isn't yet supported.  Left the updates though because when sphinx is supported it'll be less work to modify them than to do it over again

ruff.toml
```diff
@@ -3,4 +3,19 @@ exclude = [".git","__pycache__"]
 
 [lint]
 ignore= ["E741"]
```
mikewilli, line 5: I'd suggest adding `RUF005` to the ignore list, as this could lead to potential issues with numpy per [this thread](https://github.com/astral-sh/ruff/issues/2142).

This affects the changes where we change from concatenating lists to using the unpacking operator `*`, so if you adopt this suggestion we should also revert those changes.

src/mozanalysis/config.py
```diff
@@ -2,8 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from typing import List, Optional
 from dataclasses import dataclass
+from typing import List, Optional
```
mikewilli, line 6: I thought it was odd to see `List` here because the `"UP"` lint rules should be suggesting "Use `list` instead of `List` for type annotation", and in fact I see that when I run `ruff` locally on main. I went to look at why [CI is passing](https://app.circleci.com/pipelines/github/mozilla/mozanalysis/845/workflows/807f45a9-919d-44f2-b4d6-366b119a9321/jobs/2572) with this still in here and it kind of looks like ruff isn't running? Maybe I'm just missing something..

Even more strange, I see that other files (like experiment.py) have removed `List`, `Dict`, etc. in favor of `list`, `dict`, etc. as expected.

None of this is a big problem, but I'm certainly confused and a little concerned about what's going on.

ruff.toml
jaredsnyder: Ah good catch, will do

src/mozanalysis/config.py
jaredsnyder: Weirdly when I run locally ruff does not flag this and it runs fine, so maybe there is some sort of weird environment thing happening?

mikewilli: Just to clarify the behavior I'm seeing... The `UP` rule error I quoted above appears for some files but not others. I don't remember if it was different last week when I looked (I suspect not), but right now I don't see the lint error for `config.py` where I commented, but if I go to `main` where the formatting changes haven't been made yet and run `ruff` with `UP` rules, it flags other files (the ones that you've already fixed). I don't see any `# noqa` or other indications that the rule should be ignored in `config.py`, so I don't know why this is happening.

That said, it's probably not worth blocking on this odd behavior. I'll finish reviewing the rest of the PR now that I'm back.

src/mozanalysis/metrics/__init__.py
```diff
@@ -71,15 +73,15 @@ class DataSource:
     def _check_experiments_column_type(self, attribute, value):
         if value not in self.EXPERIMENT_COLUMN_TYPES:
             raise ValueError(
-                f"experiments_column_type {repr(value)} must be one of: "
-                f"{repr(self.EXPERIMENT_COLUMN_TYPES)}"
+                f"experiments_column_type {value!r} must be one of: "
+                f"{self.EXPERIMENT_COLUMN_TYPES!r}"
```
mikewilli, line 77: This change seems unnecessary to me (and possibly less obvious what it's doing), but it is a little more succinct if you're familiar with `!r`/`!s`/`!a`. What do you think? It's RUF010 just for reference.

tests/bayesian_stats/test_survival_func.py
```diff
@@ -89,7 +88,7 @@ def test_compare_branches():
     df.iloc[1::2, 0] = "test"
     df.iloc[:300, 1] = range(300)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="'df' contains null values for 'val'"):
```
mikewilli, line 91: Neat!

mikewilli (approved):
Looks good to me! Mostly converting `.format()` strings to f-strings, and sorting imports, but still nice to have.

Might be nice to get some DS eyes on this since this is mostly a DS codebase, but I'll leave that up to you. Thanks!

src/mozanalysis/metrics/__init__.py
jaredsnyder: I like it but I'm not married to it.  the `!r` format is snappy but it is a bit hard to figure out what is going on if you haven't seen it before

tests/bayesian_stats/test_survival_func.py
jaredsnyder: 😁

src/mozanalysis/config.py
jaredsnyder: Lol and now the CI is flagging the format thing, fixed

src/mozanalysis/metrics/__init__.py
mikewilli: I guess if ruff is doing it automatically for us it doesn't hurt. I think I'd lean slightly in favor of keeping it.

src/mozanalysis/config.py
mikewilli: Oh nice work! Still seems strange to me that it fixed some but not others, but glad you got it figured out.

jaredsnyder: Yeah it is super weird, had to dig through issues on Ruff's github to finally figure it out 🙃

ruff.toml
```diff
@@ -1,6 +1,22 @@
 line-length=88
 exclude = [".git","__pycache__"]
+target-version = "py310"
 
 [lint]
-ignore= ["E741"]
-select= ['E', 'F']
\ No newline at end of file
+ignore= ["E741", "RUF005"]
```
danielkberry, line 6: Why are we ignoring `E741`?

setup.py
```diff
@@ -1,17 +1,17 @@
...
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 
 tests_require = [
     "mock",
     "pytest",
     "pytest-ruff",
     "pytest-cov",
     "pytest-timeout",
+    "ruff==0.3.1",
```
danielkberry, line 14: Why are we pinning the ruff version?

src/mozanalysis/config.py
```diff
@@ -17,7 +16,7 @@ class _ConfigLoader:
     Config objects are converted into mozanalysis native types.
     """
 
-    config_collection: Optional[ConfigCollection] = None
+    config_collection: ConfigCollection | None = None
```
danielkberry, line 19: Tbh I find this and the other union type changes introduced in PEP 604 harder to read but I probably just need to get with the times. 

However, it looks like this rule, [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/), is marked unsafe for projects using runtime type annotations like Pydantic. Pydantic is in mozanalysis's dependency chain (metric-config-parser > mozilla-nimbus-schemas > pydantic). Is this rule safe for us to use?

src/mozanalysis/frequentist_stats/bootstrap.py
```diff
@@ -239,13 +237,13 @@ def compare_branches_quantiles(
         Ex: 0.2 is the 20th percentile, 0.5 is the median, etc.
     """
 
+    if quantiles_of_interest is None:
```
danielkberry, line 240: Wow, I'm impressed at the scope of the change made by this tool. Effectively, this is changing the type of that parameter from `List[float]` to `List[float] | None`, is that ok?

ruff.toml
jaredsnyder: It was ignored back when we were using black so I kept it when we transitioned to RUFF.  I can try turning it off if you think it's unnecessary

setup.py
jaredsnyder: My thinking was that it would be very annoying if the linting got out of sync because the versions were different, or if the CI was using a different version than local.  Don't know if that is a reasonable fear though

src/mozanalysis/config.py
jaredsnyder: If I'm reading the documentation right it sounds likes it's only a concern before Python 3.10, which is our minimum version for the library.

src/mozanalysis/frequentist_stats/bootstrap.py
jaredsnyder: I think the rule that's triggering is that you shouldn't have mutable types as defaults, so we could always change the default to a tuple if you don't like this fix

setup.py
danielkberry: That makes sense to me. Then we can manually update the version independently of other changes we're making. Something similar happened on my most recent Jetstream PR (where linting changes ended up being bundled with functional changes and it was difficult to review).

src/mozanalysis/frequentist_stats/bootstrap.py
danielkberry: Definitely agree that mutable types as a default is a footgun. I suppose so long as this change passes all our tests, I'm good with it.

setup.py
mikewilli: I worry that this will lead to mozanalysis being stuck on 0.3.1 forever. I think we should remove this pinning as part of the work done on dependency management that comes out of [this ADR](https://github.com/mozilla/mozanalysis/pull/216). That will allow it update like any other package, and we'll just have to be intentional about how we update packages to avoid scenarios like the one @danielkberry described.

jaredsnyder: This is a really good point, I added a comment to #202 so we don't forget

PR #215, opened 2024-03-07
add storage API to speed up data loading

### LINKED ISSUE #77: Use BQ storage api for faster download of results_table results
When `results_table` is supplied to `mozanalysis.bq.BigQueryContext.run_query()`, we could probably use the storage API to download the results table, rather than relying on the super-slow `.to_dataframe()` method currently being used. This should speed up queries by 10x from interactive users' points of view.

jaredsnyder:
Fixes #77 

I had a hard time figuring out how to best use the API and given the age of the story it might be out of date.  I ran an experiment where I locally ran the following code on this branch and on main:

```
test_target = HistoricalTarget(
    experiment_name='test_targ', # A name for the analysis, which is only used in the table names when saving results in BigQuery
    start_date='2022-01-01', # First date for the dummy enrollment period of the analysis
    num_dates_enrollment=28, # Number of days used to check for clients that satisfy conditions in Segments
    analysis_length=7, # Number of days used to calculate metrics for each client in the study
)

df = test_target.get_single_window_data(
    bq_context=bq_context,
    metric_list=[active_hours, uri_count, search_count],
    target_list=[weekday_regular_us]
)
```

I used the jupyter `%%timeit` magic to measure how fast.  On main it ran in `37.4 s ± 8.17 s per loop (mean ± std. dev. of 7 runs, 1 loop each)` while on the feature branch it ran in `31.3 s ± 1.4 s per loop (mean ± std. dev. of 7 runs, 1 loop each)`.  Not very scientific, especially given the std deviation on the main, but it looks like the improvement is fairly minor in this case.

### DIFF
--- a/src/mozanalysis/bq.py
+++ b/src/mozanalysis/bq.py
@@ -6,6 +6,8 @@ import re
 
 from google.api_core.exceptions import Conflict
 from google.cloud import bigquery
+from google.cloud.bigquery_storage import BigQueryReadClient
+
 
 logger = logging.getLogger(__name__)
 
@@ -38,6 +40,7 @@ class BigQueryContext:
         self.dataset_id = dataset_id
         self.project_id = project_id
         self.client = bigquery.Client(project=project_id)
+        self.read_client = BigQueryReadClient()
 
     def run_query(self, sql, results_table=None, replace_tables=False):
         """Run a query and return the result.
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -276,7 +276,9 @@ class Experiment:
             "_".join([last_date_full_data, self.experiment_slug, hash_ish(metrics_sql)])
         )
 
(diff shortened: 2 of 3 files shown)


m-d-bowerman:
I believe the `to_dataframe` method already does this under the hood, based on the `create_bqstorage_client` [parameter](https://github.com/googleapis/python-bigquery/blob/08762fbf3289622c2b948da919ae49567d0378d5/google/cloud/bigquery/table.py#L2087-L2095) that defaults to true.

jaredsnyder:
> I believe the `to_dataframe` method already does this under the hood, based on the `create_bqstorage_client` [parameter](https://github.com/googleapis/python-bigquery/blob/08762fbf3289622c2b948da919ae49567d0378d5/google/cloud/bigquery/table.py#L2087-L2095) that defaults to true. I guess passing the client explicitly does avoid creating that client each time the `to_dataframe` method is called.

Lol think we can just close this PR and the issue then

PR #216, opened 2024-03-08
Added ADR file to evaluate methods for better dependency resolution

jaredsnyder:
Addresses #202 

Evaluates the pros and cons of using pip-tools and poetry for better dependency resolution.

### DIFF
--- a/docs/adrs/adr-0001-dependency_management.md
+++ b/docs/adrs/adr-0001-dependency_management.md
@@ -1,7 +1,7 @@
 # Tooling for Better Dependency Management
 
-* Status: 
-* Deciders: 
+* Status: Pending
+* Deciders: jsnyder, mikewilli
 * Date: 2024-03-08
 
 
@@ -19,8 +19,8 @@ Currently dependencies are very loosly defined in the `setup.py`. This can lead 
 ### 1. pip-tools to generate requirements.txt
 Jetstream currently uses this process to manage dependencies.  It involves uses pip-tools to generate a requirements.txt and requirements.in file from the setup.py. [dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) is a built-in github tool that is used to periodically check for dependency updates and create pull requests to implement them.  [Here](https://github.com/mozilla/jetstream/pull/2017) is an example of such a pull request.
 
+* : Dependabot already used in Jetstream
 * +: Jetstream uses it already so there is an existing base of experties in the data team
-* +: Dependabot is a built-in github tool that has additional uses like checking for security vulnerabilities
 * +: Dependencies can be automatically updated with a script like the [update_deps](https://github.com/mozilla/jetstream/blob/main/script/update_deps) script in Jetstream
 * -: Still using pip's dependency resolver, which is fast but can have issues with incompatible versions.  This may be mitigated or even solved by the way pip-tools pins dependencies.
 * -: Jetstream doesn't use `pyproject.toml` for configuring dependencies, though it does seem to be supported by `pip-tools`
@@ -28,6 +28,7 @@ Jetstream currently uses this process to manage dependencies.  It involves uses 
 ### 2. poetry
 (poetry)[https://python-poetry.org/] is a popular tool for dependency managment.  When starting development with a poetry project, one uses the `poetry install` command to create a virtual environment for the project and install the package and its dependencies.  New dependencies are added with the `poetry add` command.  poetry (and other tools) are configured in the `pyproject.toml`.
 
+* : Can add Dependabot, which is alrady built-in to github tool that has additional uses like checking for security vulnerabilities
 * +: Very reliable dependency resolution
 * +: Uses `pyproject.toml`, which is the new standard and is nice for consolidating other config files
 * +: Also acts as a virtualenv manager
@@ -39,3 +40,5 @@ Jetstream currently uses this process to manage dependencies.  It involves uses 
 * -: would need to update tox to use poetry to build and publish package
 
 ## Decision Outcome
+
+The pip-tools method wins out because it is familiar and simple.  It would be nice to use the PR that adds this functionality to move as much configuation as possible in to a `pyproject.toml` file.
--- a/docs/adrs/adr-0001-dependency_management.md
+++ b/docs/adrs/adr-0001-dependency_management.md
@@ -39,6 +39,15 @@ Jetstream currently uses this process to manage dependencies.  It involves uses 
 * -: Not wildely use in the data team
 * -: would need to update tox to use poetry to build and publish package
 
+### 3. mamba/conda
+(mamba)[https://mamba.readthedocs.io/en/latest/] is a fast, robust, and cross-platform package manager. It is fully compatible with conda, which is a common package manager in Python, and can be a drop-in replacement for conda. Mamba creates a virtual environment that is isolated from each other.
+
+* Creates isolated virtual environment that encapsulates all the packages in its own designated space
+* Can assign a specific Python version to the given virtual environment
+* Dependency resoluion is fast and optimized for parallelism
+* Uses an `environment.yml` to specify dependencies
+* Can consume `pyproject.toml` when specifying packages
+
 ## Decision Outcome
 
 The pip-tools method wins out because it is familiar and simple.  It would be nice to use the PR that adds this functionality to move as much configuation as possible in to a `pyproject.toml` file.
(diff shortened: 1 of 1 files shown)


docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Options
+
+### 1. pip-tools to generate requirements.txt
+Jetstream currently uses this process to manage dependencies.  It involves uses pip-tools to generate a requirements.txt and requirements.in file from the setup.py. [dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) is a built-in github tool that is used to periodically check for dependency updates and create pull requests to implement them.  [Here](https://github.com/mozilla/jetstream/pull/2017) is an example of such a pull request.
+
+* +: Jetstream uses it already so there is an existing base of experties in the data team
+* +: Dependabot is a built-in github tool that has additional uses like checking for security vulnerabilities
+* -: Still using pip's dependency resolver, which is fast but can have issues with incompatible versions.  This may be mitigated or even solved by the way pip-tools pins dependencies.
+* -: Jetstream doesn't use `pyproject.toml` for configuring dependencies, though it does seem to be supported by `pip-tools`
```
mikewilli, line 26: Yea I think this is more a relic of Jetstream's history than an intentional or necessary decision, and we could change to `pyproject.toml` if we want. Worth calling out here though.

jaredsnyder: Based on some online sleuthing it seems like the `pyproject.toml` support is relatively new too

docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,40 @@
...
+
+
+## Context and Problem Statement
+Currently dependencies are very loosly defined in the `setup.py`. This can lead to issues when installing the package, an example of which is detailes [in this issue](https://github.com/mozilla/mozanalysis/issues/202).  We need to find a way to more clearly define dependencies in a way that will not be disruptive for developers and users.
+
+## Decision Drivers
+- Minimize the potential for incompatible dependencies
+- Have guardrails/automation so that changes to dependencies made by developers will easily be reflected in the PRs they make
+- Ideally no disruption to user workflow
+- Minimal distruption to developer workflow
```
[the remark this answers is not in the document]
mikewilli: That makes sense, though I think it's the same for both options here. Important criteria vs the existing situation though, thanks.

docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Options
+
+### 1. pip-tools to generate requirements.txt
+Jetstream currently uses this process to manage dependencies.  It involves uses pip-tools to generate a requirements.txt and requirements.in file from the setup.py. [dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/) is a built-in github tool that is used to periodically check for dependency updates and create pull requests to implement them.  [Here](https://github.com/mozilla/jetstream/pull/2017) is an example of such a pull request.
+
+* +: Jetstream uses it already so there is an existing base of experties in the data team
+* +: Dependabot is a built-in github tool that has additional uses like checking for security vulnerabilities
```
mikewilli, line 23: Fairly certain that dependabot works with poetry as well, just fyi.

docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,41 @@
...
+
+* +: Very reliable dependency resolution
+* +: Uses `pyproject.toml`, which is the new standard and is nice for consolidating other config files
+* +: Also acts as a virtualenv manager
+* +: No changes to user workflow
+* +: When a new dependency is added with `poetry add` the `poetry.lock` and `pyproject.toml` files are automatically updated so there minimal risk of a new dependency not being accounted for
+* -: Dependency resolution is slow, when you add a new package it can take several minutes to figure it out
+* -: Developers would have to modify their workflow to accomodate poetry's virtualenv management
+* -: Not wildely use in the data team
+* -: would need to update tox to use poetry to build and publish package
```
mikewilli, line 39: I wonder if it makes sense to replace tox if we decide to use poetry?

jaredsnyder: Yeah one annoying thing about poetry is that when you use it with nice CI tools like tox you get the matryoshka doll of virtualenvs.  I'm not sure what the alternative to tox that plays nicely with poetry and all the other moving pieces is though.

docs/adrs/adr-0001-dependency_management.md
jaredsnyder: Ah neato, guess I can remove this bullet.  I think it does use pip under the hood with poetry-based libraries but I don't think it matters 
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

docs/adrs/adr-0001-dependency_management.md
mikewilli: Well the goal of both of them is to have a consistent environment, so I was thinking that maybe with poetry you don't need tox.

docs/adrs/adr-0001-dependency_management.md
mikewilli: Yea Experimenter uses both poetry and dependabot, so should be doable. Worth including dependabot in both options I think.

docs/adrs/adr-0001-dependency_management.md
jaredsnyder: Yeah I guess it wouldn't be that hard to take the commands from the `tox.ini` and move them into the circleci config if we were using poetry

teonbrooks: as a suggestion, have we considered using mamba/conda to generate a virtual environment with all the requirements. it's pretty robust with creating an `environment.yml` and it is quite fast and performant

jaredsnyder: That'd be worth looking into, would you mind adding a section to the ADR for mamba/conda?

docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,53 @@
...
+
+### 3. mamba/conda
+(mamba)[https://mamba.readthedocs.io/en/latest/] is a fast, robust, and cross-platform package manager. It is fully compatible with conda, which is a common package manager in Python, and can be a drop-in replacement for conda. Mamba creates a virtual environment that is isolated from each other.
+
+* Creates isolated virtual environment that encapsulates all the packages in its own designated space
+* Can assign a specific Python version to the given virtual environment
```
jaredsnyder, line 46: My biggest concern with conda is that all our docs assume people are using pip to install, and in my experience using pip install in conda envs can lead to big headaches.  Is that still the case

docs/adrs/adr-0001-dependency_management.md
```diff
@@ -0,0 +1,53 @@
...
+* Dependency resoluion is fast and optimized for parallelism
+* Uses an `environment.yml` to specify dependencies
+* Can consume `pyproject.toml` when specifying packages
```
jaredsnyder, line 49: A `-` we should note is that we'd have to come up with our own script to do the automatic update based on the venv like Jetstream does

mikewilli: A couple other negatives:
- we note that poetry is not widely used in the data team, but is conda used in any mozilla data repos? I only regularly work with around a dozen repos and those use pip-tools and poetry but not conda/mamba
- I believe the storage size of a mozanalysis environment would drastically increase, at least from what I remember of working with conda. Doesn't it come with a lot of libraries pre-packaged?
- the isolated virtual environment comes with potential for confusion over system libraries vs conda libraries as developers move between projects

I'm intentionally searching for negatives here to ensure we're considering everything (I think @teonbrooks 
covered the positives pretty well already). I have used conda in the past and found a lot to like about it, but also some frustrations. That said, I wasn't aware of poetry at the time -- what are the advantages vs poetry for our situation?

teonbrooks: @mikewilli to answer your question, conda environments do not come with any packages pre-installed, it comes with the default or specified python version and that's pretty much it. 

with Conda there is some confusion with the Anaconda and the bundled solution they offer. They do use conda as a package manager but it also comes with a host of pre-configured packages.

jaredsnyder: Thanks for the update Teon.  After looking it over I still think pip-tools is the best option just because of familiarity.  In the end I think it's pretty similar to the poetry vs pip-tools decision.  It looks like the big advantage of conda is a speedier dependency resolver and the built-in virtualenvs, and I don't think that outweighs the cost of having to figure out how to rework the Jetstream scripts to use it.

PR #217, opened 2024-03-12
docs: update docstrings for custom_exposure_query and custom_enrollments_query

### LINKED ISSUE #207: Clarify custom_enrollments_query docs
[These docs](https://mozilla.github.io/mozanalysis/api/experiment.html#mozanalysis.experiment.Experiment.get_single_window_data) are a bit confusing and incomplete for `custom_enrollments_query`:
```
A full SQL query to be used in the main query:

`WITH raw_enrollments AS ({custom_enrollments_query})`

N.B. this query’s results must be uniquely keyed by (client_id, branch), or else your results will be subtly wrong.
```
This text (or similar) is copied for multiple functions within this package, any updates should be made to all relevant function docs.

We should make the following updates to clarify:
- make it more clear that the `custom_enrollments_query` is only the part inside the parens (i.e., should _not_ include "WITH raw_enrollments AS"
- per #93, the query needs to produce columns client_id, branch, enrollment_date, and num_enrollment_events
- per #93, it should have no more than one row per (client_id, branch)

jaredsnyder:
Fixes #207

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -181,20 +181,19 @@ class Experiment:
                 Specifies the query type to use to get the experiment's
                 enrollments, unless overridden by
                 ``custom_enrollments_query``.
-            custom_enrollments_query (str): A full SQL query to be used
-                in the main query::
+            custom_enrollments_query (str): A full SQL query that
+                will generate the `enrollments` common table expression
+                used in the main query. The query must produce the columns
+                `client_id`, `branch`, `enrollment_date`, and `num_enrolled_events`.
 
-                    WITH raw_enrollments AS ({custom_enrollments_query})
-
-                N.B. this query's results must be uniquely keyed by
+                WARNING: this query's results must be uniquely keyed by
                 (client_id, branch), or else your results will be subtly
                 wrong.
 
-            custom_exposure_query (str): A full SQL query to be used in the main
-                query::
-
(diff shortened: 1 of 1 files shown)


src/mozanalysis/experiment.py
```diff
@@ -925,9 +926,7 @@ def _build_metrics_query_bits(
                 """    LEFT JOIN (
         {query}
         ) ds_{i} USING (client_id, branch, analysis_window_start, analysis_window_end)
-                """.format(
-                    query=query_for_metrics, i=i
-                )
+                """.format(query=query_for_metrics, i=i)
```
mikewilli, line 929: No f-strings in this file?

mikewilli (approved):
LGTM thanks! Surprised ruff doesn't care about the `.format()` calls in this file, but since this PR is about doc updates I think that's a problem for another time.

src/mozanalysis/experiment.py
jaredsnyder: I'm still using the old ruff rules because I'm still waiting on merging the ruff branch.  Waiting for a final approval from one of the DS folks, I'll ask about it today

mikewilli: Whoops :facepalm: that makes sense! Glad it's not another strange case of some files skipping some lint rules.

PR #220, opened 2024-03-18
Verify target_list and metric_list sources match

### LINKED ISSUE #173: Add checks in sizing for disparate datasets
Currently, no warning is raised when Metrics and Segments are using data sources that rely on different applications' datasets. A check should be added in `sizing.py` for a mismatch between data sources.

jaredsnyder:
Fixes #173 

- Add `app_name` attribute to Metric, Segement, etc classes so that sources from metric-hub can be checked
- exceptions raised for metric_list with multiple sources, target_list with multiple sources and when the sources between the two don't match
- Add warning for causes where there are metrics or segements not from metric-hub, and a metric has all 0s (which would indicate no support for the metric and therefore a mismatch between target and metric sources)

### DIFF
--- a/src/mozanalysis/sizing.py
+++ b/src/mozanalysis/sizing.py
@@ -3,9 +3,9 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from __future__ import annotations
 
+import warnings
 from datetime import date
 from typing import TYPE_CHECKING
-import warnings
 
 import attr
 
--- a/src/mozanalysis/sizing.py
+++ b/src/mozanalysis/sizing.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from datetime import date
 from typing import TYPE_CHECKING
+import warnings
 
 import attr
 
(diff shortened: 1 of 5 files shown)


src/mozanalysis/sizing.py
```diff
@@ -162,6 +163,27 @@ def get_single_window_data(
                 "Either custom_target_query or target_list must be provided"
             )
 
+        # validate metric_list and target_list are from the same source
+        # filter out el.app_name is None, there are cases where metric-hub
+        # was not used and it's up to users to ensure the sources match
+        metric_list_sources = {el.app_name for el in metric_list if el.app_name}
+        target_list_sources = {el.app_name for el in target_list if el.app_name}
+
+        if len(metric_list_sources) > 1:
```
m-d-bowerman, line 172: We should probably raise warnings here, rather than errors, as I could see some cases where multiple app_names might be used -- for instance, joins to the metrics in multi-product

src/mozanalysis/sizing.py
```diff
@@ -226,10 +248,26 @@ def get_single_window_data(
...
+            len(target_list_sources) < len(target_list)
+        ):
+            # case where there is a custom definition
+            # so sources have not been checked
+            for metric_obj in metric_list:
+                if all(output[metric_obj.name] == 0):
+                    raise warnings.warn(
+                        (
+                            f"Metric {metric_obj.name} is all 0, which may indicate"
+                            + " segements and metric do not have a common source"
```
m-d-bowerman, line 265: I like this check, and I think it should always run -- for example, we could have two Firefox Desktop metrics, but one uses a Glean client_id and one uses a legacy client_id, and this check would fail.

jaredsnyder: Fixed

src/mozanalysis/sizing.py
jaredsnyder: That makes sense, fixed and updated the tests which had to modified to accommodate this

m-d-bowerman:
The other thing that occurred to me is that using "source" in the warnings might be ambiguous -- users might take it to mean multiple `DataSource`s are causing the warnings. I think using "app" might be more informative (e.g. `segments and metric do not have a common app`)

jaredsnyder:
> The other thing that occurred to me is that using "source" in the warnings might be ambiguous -- users might take it to mean multiple `DataSource`s are causing the warnings. I think using "app" might be more informative (e.g. `segments and metric do not have a common app`)

Good catch, fixed it

m-d-bowerman (approved):
lgtm

PR #226, opened 2024-05-10
Linear models

danielkberry:


src/mozanalysis/frequentist_stats/linear_models.py
```diff
@@ -0,0 +1,412 @@
...
+    - alphas (list[float]): the desired confidence levels
+
+    Returns:
+    - out (dict[str, pd.Series]): a dictionary keyed by branch label with a pandas
+    series as value. The format of the series is described above in
+    `summarize_one_branch`
+    """
+    return {
+        b: summarize_one_branch(data.loc[(branches == b).values], alphas)
+        for b in branch_list
```
jaredsnyder, line 80: In most cases with `branch_list == branches.unique()`?  If so, maybe make it an optional parameter that defaults to `branches.unique()`

src/mozanalysis/frequentist_stats/linear_models.py
```diff
@@ -0,0 +1,412 @@
...
+    Inferences on `\beta_t1` are inferences of the average treatment effect (ATE) of
+    branch t1. That is, the confidence interval for `\beta_t1` is the confidence
+    interval for the ATE of branch t1. Similarly for t2 and `\beta_t2`.
+
+    We can incorporate a single (generally pre-experiment) covariate. Adding a covariate
+    of `Ypre` to the above, we'll build the following formula:
+    `Y ~ C(branch, Treatment(reference='control')) + Ypre`
+    which will fit the following linear model:
+
+    `Y_i = \beta_0 + \beta_t1*I(branch == 't1') + \beta_t2*I(branch == 't2') + Ypre_i`
```
jaredsnyder, line 107: Why doesn't the covariate get a parameter?  Why can you only have one?

src/mozanalysis/frequentist_stats/linear_models.py
```diff
@@ -0,0 +1,412 @@
...
+    - ref_branch (str): the name of the reference branch
+    - covariate (Optional[str]): the name of a covariate to include in the model.
+
+    Returns:
+    - formula (str): the R-style formula, to be passed to statsmodels's formula API.
+
+    """
+    formula = f"{target} ~ C(branch, Treatment(reference='{ref_branch}'))"
+    if covariate is not None:
+        formula += f" + {covariate}"
```
jaredsnyder, line 123: If there are forbidden characters for the string, maybe have a check that `target`, `ref_branch` and `covariate` don't contain those strings?  Seems kind of insane someone would put a tilde in a variable names but couldn't hurt 🤷

jaredsnyder: Might make more sense to put that check in `summarize_joint`

tests/frequentist_stats/test_linear_models.py
```diff
@@ -0,0 +1,510 @@
...
+    assert expected == actual
+
+    expected = "days_of_use ~ C(branch, Treatment(reference='treatment-a'))"
+    actual = mafslm._make_formula("days_of_use", "treatment-a")
+
+    assert expected == actual
+
+    expected = (
+        "active_hours ~ C(branch, Treatment(reference='treatment-a'))"
+        " + active_hours_pre"
```
jaredsnyder, line 132: nit picking but it'd be appropriate to ignore the line length rule here imo

src/mozanalysis/frequentist_stats/linear_models.py
danielkberry: 1) because I forgot. Edited the formula to add it in. 2) you don't have to have only one, I'm thinking we start with 1 (the pre-treatment version of the metric being studied) for simplicity, and can expand from there. I'd like to see us get to many, but that's much more complex.

tests/frequentist_stats/test_linear_models.py
```diff
@@ -0,0 +1,510 @@
...
+    branch = ["control"] * 100
+    df_in = pd.DataFrame({"y": y, "branch": branch, "y2": y2})
+    df_actual = mafslm._make_model_df(
+        df_in, "y", covariate_col_label="y2", threshold_quantile=0.95
+    )
+    df_expected = pd.DataFrame(
+        {
+            "branch": branch,
+            "y": [float(_y) for _y in y[:-5] + [np.quantile(y, 0.95)] * 5],
+            "y2": [float(_y) for _y in y2[:-5] + [np.quantile(y2, 0.95)] * 5],
```
jaredsnyder, line 444: Why isn't this just list(np.quantile(y, .95)) ?

tests/frequentist_stats/test_linear_models.py
```diff
@@ -0,0 +1,510 @@
...
+from statsmodels.stats.weightstats import CompareMeans
+
+
+def _make_test_model():
+    ref_branch = "treatment-a"
+    alphas = [0.01, 0.05]
+    branches = ["control"] * 100 + ["treatment-a"] * 100 + ["treatment-b"] * 100
+    searches = list(range(100)) + list(range(100, 200)) + list(range(200, 300))
+    model_df = pd.DataFrame({"search_count": searches, "branch": branches})
+    formula = mafslm._make_formula("search_count", ref_branch)
```
jaredsnyder, line 16: Write out the formula as a string here rather than relying on `_make_formula`

jaredsnyder (changes requested):
Great stuff here, the test coverage is excellent and the code is well written and documented.  Mostly requests for clarification and nit picks.  

I ran the tests locally and it is failing on `test_summarize_univariate` and `test_summarize_joint`.  In both cases the error is:


```

if alpha <= 0 or alpha >= 1:
TypeError: '<=' not supported between instances of 'str' and 'int'

```

src/mozanalysis/frequentist_stats/linear_models.py
danielkberry: I added a function to do this.

tests/frequentist_stats/test_linear_models.py
danielkberry: TIL you could do that.

src/mozanalysis/frequentist_stats/linear_models.py
```diff
@@ -0,0 +1,412 @@
...
+        abs_uplifts = _extract_relative_uplifts(
+            results, branch, ref_branch_label, alphas
+        )
+
+        output[branch] = pd.concat([rel_uplifts, abs_uplifts])
+
+    return output
+
+
+def _make_model_df(
```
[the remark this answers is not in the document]
danielkberry: I was initially thinking that this should be private, but upon further review I agree that's probably not necessary

tests/frequentist_stats/test_linear_models.py
danielkberry: Because the top 5 elements are clipped to the 95th percentile

tests/frequentist_stats/test_linear_models.py
```diff
@@ -0,0 +1,510 @@
...
+    )
+
+    bootstrap_results = mafsb.compare_branches(
+        model_df, column_label, ref_branch_label=ref_branch
+    )
+
+    # control vs treatment-a
+    out = mafslm._extract_relative_uplifts(results, "control", ref_branch, alphas)
+
+    expected_mean = -0.1 / 2
```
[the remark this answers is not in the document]
danielkberry: Nice catch

tests/frequentist_stats/test_linear_models.py
danielkberry: Another way to write this might be `np.clip(y, 0, np.quantile(y, 0.95)).astype(float)`

tests/frequentist_stats/test_linear_models.py
```diff
@@ -0,0 +1,510 @@
...
+        ("rel_uplift", "0.975"),
+        ("rel_uplift", "0.995"),
+        ("rel_uplift", "exp"),
+    ]
+
+    for branch in comparative_branches:
+        assert sorted(out["comparative"][branch].index) == expected_index
+
+
+def test_fit_model():
```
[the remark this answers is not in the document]
danielkberry: Done

tests/frequentist_stats/test_linear_models.py
jaredsnyder: AHHH okay I get it

src/mozanalysis/frequentist_stats/linear_models.py
danielkberry: Added!

PR #228, opened 2024-05-16
Improving dependency management as per ADR 0001

### LINKED ISSUE #202: Better dependency resolution

jaredsnyder:
This is a PR to improve the dependency management by adding a `requirements.txt` file autogenerated by a script via `pip-tools`.  More details in the ADR 0001 doc in `docs/adrs`.  Added a `pyproject.toml` to consolidate a lot of the config too.

Non-exhaustive list of things I need reviewers to double-check:
- Make sure nothing from the `setup.cfg` and `setup.py` that we need got left out of the `pyproject.toml`.  There was some stuff that seemed irrelevant (IE all the `isort` config from the `setup.cfg` which we don't seem to be using) so make sure that is actually irrelevant.
- that tox works locally for you
- that the auto-generated version is working as expected
- using the requirements.txt works as expected
- check to make sure I set up dependabot correctly

Also questions:
-  Do we want to use this PR to clean up the `tox.ini`?  It looks significantly different than jetstreams.  If so I'll need some help understanding the difference probably.
- Do we need to build a universal wheel given we don't support python2?

Fixes #202


README.md
```diff
@@ -14,9 +14,19 @@ Online documentation is available at https://mozilla.github.io/mozanalysis/
 pip install mozanalysis
 ```
 
+## Local Installation
+```bash
+# Create and activate a python virtual environment.
+python3 -m venv venv/
+source venv/bin/activate
+pip install -r requirements.txt
```
danielkberry, line 22: We need a `pip install -e .` line following this. The instructions as written only install the dependencies, not the actual package. And yeah... Jetstream's instructions are wrong too.

README.md
```diff
@@ -14,9 +14,19 @@ Online documentation is available at https://mozilla.github.io/mozanalysis/
...
+```
+
 ## Development
 Linting and Formatting are done with Ruff.
 
+Update the `requirements.txt` file using the script `script/update_deps`
+
```
danielkberry, line 29: What's the actual procedure to update a dependency? Do you manually install the dependency into your environment and then run `update_deps`?

tox.ini
```diff
@@ -5,23 +5,22 @@
 [tox]
```
danielkberry, line 5: So tox did not install as part of installing mozanalysis. Is this expected? Are users intended to have their own tox version?

danielkberry (changes requested):
Just one quick change. Otherwise it LGTM. Tested on a fresh vertex workbench.

README.md
jaredsnyder: Lol that's what I get for copying and pasting

tox.ini
jaredsnyder: That's a good question, maybe we want to add a `dev` optional dependency list?  pip-tools is in the same boat

README.md
jaredsnyder: added to the README

README.md
jaredsnyder: Added to the README

tox.ini
jaredsnyder: Added optional dev dependencies and update the requirements files to use those

jaredsnyder: Considered adding separate requirements files for dev dependencies but it felt a little too much so I didn't

README.md
danielkberry: Good thing I asked lol

tox.ini
jaredsnyder: Yeah that's basically it.  It might make sense to just combine it into one because the only real use case for separating them that I can think of is having the CI using `testing` while developers use `dev`, but we're generating the requirements files from `dev` and that's what the CI is using to setup so 🤷

jaredsnyder: I guess the counterpoint is that it does provide a little self-documentation to keep them separate, and because the toml file lets us reference the testing deps in dev we're not introducing any potential for them to get out of sync

danielkberry: I'm good to keep them separate. Is updating the CI to use the `testing` requirements in scope for this PR or should we wait on that?

jaredsnyder: I don't think we need to do it at all but I would like to hear Anna or Mike's opinion honestly.  As I see it, tox is installing in the venv that is created by tox, which could cause problems maybe but it doesn't seem to be so 🤷

README.md
```diff
@@ -14,9 +14,21 @@ Online documentation is available at https://mozilla.github.io/mozanalysis/
...
+source venv/bin/activate
+pip install -r requirements.txt
+pip install -e .
+```
+
 ## Development
+After installing the dependencies as described above, install the package and additional development dependencies with `pip install -e ".[dev]"`
 Linting and Formatting are done with Ruff.
 
+When adding new dependencies, add them to the `pyproject.toml` `dependencies` list.  Then update the `requirements.txt` file using the script `script/update_deps`.  [`pip-compile` is called](https://pypi.org/project/pip-tools/) in this script, which uses the specified dependencies to create the `requirements.txt` file.
```
mikewilli, line 30: Ideally all dependencies are kept up to date, but in the event that doesn't happen (like with Jetstream, which we go through and update periodically, but sometimes there can be a dozen or two dependencies that have available updates), should we document a way for someone to update a single dependency?

In Jetstream, that means manually editing `requirements.in` with the new version of the package you want to use. Not sure if you have a better way, but could add that here too.

pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+    "C4", # flake8-comprehensions
+    "I", # isort
+    "SIM", # flake8-simplify
+    "TCH", # flake8-type-checking
+    "TID", # flake8-tidy-imports
+    "Q", # flake8-quotes
+    "UP", # pyupgrade
+    "PT", # flake8-pytest-style
+    "RUF", # Ruff-specific rules
+]
```
mikewilli, line 64: Hooray for consolidation!

mikewilli (changes requested):
Changes look great! Pretty sure this needs some updates to circle's config.yml though, particularly since `setup.py` is being deleted here and is referenced in the deploy step. I think that's the only part that needs updating, but take a quick look if you haven't already in case I missed something.

If you want to test the CI deploy step more than just local testing but without merging and tagging, you can edit the triggers so that it will deploy on the branch without a tag, and then change it back before merging. If you go this route, it's probably a good idea to also change `twine upload` to just `twine check` (or at least add `--repository-url https://test.pypi.org/` to `twine upload`) so that we don't clutter pypi.

mikewilli:
> Do we want to use this PR to clean up the tox.ini? It looks significantly different than jetstreams. If so I'll need some help understanding the difference probably.

I don't think the differences are a concern. I don't see any differences that are important to change, and changes would probably be made to the jetstream side if anything (e.g., incorporate rust).

> Do we need to build a universal wheel given we don't support python2?

No, not as far as I understand at least.

tox.ini
mikewilli: I think installing tox inside tox is probably not going to cause problems, but definitely feels like it would indicate a misconfiguration or misuse somewhere (we shouldn't be running tox inside tox, at the very least).

However, `tox` is only in the `dev` list, which shouldn't be needed for CI, so I think using `testing` inside CI makes sense. I don't think that would be a complicated change, but CI changes are often annoying so if you'd rather tackle it in a follow-up that's fine with me. Also fine to put it on the backlog as it doesn't seem like a high priority or high value change either.

README.md
jaredsnyder: I would say if someone needs to do that they should specify a minimum version for whatever dependency they want to update in the `pyproject.toml` and then update with the standard command.  Putting the minimum version provides some self documentation and I think there is value in having to update the dependencies "all the way" rather than doing a kind of bandaid.  What do you think?

tox.ini
jaredsnyder: So I ended up just putting the dev requirements in their own file.  It was an easy update to `update_deps`, now the CI can use the normal requirements.txt and it won't have tox. let me know what you think.

README.md
mikewilli, line 30: Oh one other thing -- consider adding a note to re-run `pip install -r requirements.txt` after regenerating it. Easy to forget (at least for me).

tox.ini
mikewilli: I'm not opposed, but does there have to be so much overlap? It would be nice if the dev one could have *just* the dev deps (but maybe the majority of stuff in both of them is sub-dependencies?).

Does the README need to be updated to account for this change?

mikewilli (approved):
Two small suggestions but otherwise LGTM! :tada:

README.md
jaredsnyder: Agreed!

jaredsnyder: Good call, pushed an update

tox.ini
jaredsnyder: So I looked through the docs and we can use the `-c` flag so that the `requirements-dev.txt` will be constrained by the `requirements.txt`. It's still overlapping but at least we don't have to worry about conflicts

jaredsnyder: And the documentation should reflect this change, I distinguish between the two files in the relevant sections

jaredsnyder: relevant docs on the c flag: https://github.com/jazzband/pip-tools?tab=readme-ov-file#workflow-for-layered-requirements

README.md
```diff
@@ -14,9 +14,22 @@ Online documentation is available at https://mozilla.github.io/mozanalysis/
...
+pip install -r requirements.txt
+pip install -e .
+```
+
 ## Development
+Follow the steps above, substituting `requirements-dev.txt` for the requirements file. Install the package and additional development dependencies with `pip install -e ".[dev]"`
+
 Linting and Formatting are done with Ruff.
 
+When adding new dependencies, add them to the `pyproject.toml` `dependencies` list.  Then update the `requirements.txt` file using the script `script/update_deps`.  [`pip-compile` is called](https://pypi.org/project/pip-tools/) in this script, which uses the specified dependencies to create the `requirements.txt` file.  Finally, you'll want to update the requirements in your virtual env by running `pip install -r requirements.txt`
```
mikewilli, line 31:
```suggestion
When adding new dependencies, add them to the `pyproject.toml` `dependencies` list.  Then generate an updated `requirements.txt` file using the script `script/update_deps`.  [`pip-compile` is called](https://pypi.org/project/pip-tools/) in this script, which uses the specified dependencies to create the `requirements.txt` file.  Finally, you'll want to update the requirements in your virtual env by running `pip install -r requirements.txt`
```

jaredsnyder: Just pushed this change

danielkberry (approved):
LGTM

README.md
```diff
@@ -14,22 +14,31 @@ Online documentation is available at https://mozilla.github.io/mozanalysis/
...
+python3 -m venv venv/
+source venv/bin/activate
+pip install -r requirements.txt
+pip install -e . --no-dependencies
+```
+
 ## Development
 Linting and Formatting are done with Ruff.
 
+When adding new dependencies, add them to the `pyproject.toml` `dependencies` list.  Then generate updated requirmenets files using the script `script/update_deps`.  [`pip-compile` is called](https://pypi.org/project/pip-tools/) in this script, which uses the specified dependencies to create the `requirements.txt` and `requirements-dev.txt` files.  Finally, you'll want to update the requirements in your virtual env by running `pip install -r requirements-dev.txt`
```
mikewilli, line 32: oop
```suggestion
When adding new dependencies, add them to the `pyproject.toml` `dependencies` list.  Then generate updated requirements files using the script `script/update_deps`.  [`pip-compile` is called](https://pypi.org/project/pip-tools/) in this script, which uses the specified dependencies to create the `requirements.txt` and `requirements-dev.txt` files.  Finally, you'll want to update the requirements in your virtual env by running `pip install -r requirements-dev.txt`
```

PR #233, opened 2024-05-21
remove configloader helper objects

### LINKED ISSUE #154: Update docs to reflect metrics migration
[Docs like these](https://mozilla.github.io/mozanalysis/api/metrics/desktop.html) are empty even though the namespaces are not.

jaredsnyder:
Fixes #154

Removes deprecated metrics and segments subpackages.  Important classes definied in the `__init__.py` of these were moved to `modules.py` and `segments.py` files.

I haven't worked with sphinx much so the main thing I'm worried about is that I broke it by deleting those `.rst` files


tests/test_experiment.py
```diff
@@ -17,6 +12,44 @@
 from mozanalysis.segments import Segment, SegmentDataSource
 
 
+@pytest.fixture()
+def desktop_metrics():
```
danielkberry, line 16: Seems like this functionality is defined in a bunch of test files. How about placing it in a single file and then all the test files can import?

tests/test_experiment.py
```diff
@@ -319,15 +352,22 @@ def test_megaquery_not_detectably_malformed():
...
 
-def test_segments_megaquery_not_detectably_malformed():
+def test_segments_megaquery_not_detectably_malformed(desktop_metrics):
+    desktop_segments = [
+        ConfigLoader.get_segment("regular_users_v3", "firefox_desktop"),
+        ConfigLoader.get_segment("new_or_resurrected_v3", "firefox_desktop"),
+        ConfigLoader.get_segment("weekday_regular_v1", "firefox_desktop"),
+        ConfigLoader.get_segment("allweek_regular_v1", "firefox_desktop"),
+        ConfigLoader.get_segment("new_unique_profiles", "firefox_desktop"),
+    ]
```
danielkberry, line 370: Same question here: can we define test segments in that single file?

tests/test_experiment.py
```diff
@@ -339,22 +379,22 @@ def test_segments_megaquery_not_detectably_malformed():
 
     enrollments_sql = exp.build_enrollments_query(
         time_limits=tl,
-        segment_list=[s for s in msd.__dict__.values() if isinstance(s, Segment)],
+        segment_list=desktop_segments,
```
danielkberry, line 382: The new approach seems much cleaner

tests/test_metric_libraries.py
```diff
@@ -1,51 +1,102 @@
...
+        other_datasources += [
+            ConfigLoader.get_data_source("baseline", platform),
+            ConfigLoader.get_data_source("events", platform),
+            ConfigLoader.get_data_source("metrics", platform),
+        ]
+
+    return desktop_datasources + other_datasources
 
 
 def test_sql_not_detectably_malformed(included_metrics, included_datasources):
```
danielkberry, line 91: Does this test mean that mozanalysis is testing the validity of definitions from metric-hub? Or is this testing something about mozanalysis's sql generation? I'd think the former should be the responsibility of metric-hub to test, though they're only testing on merges, not continuously.

danielkberry (changes requested):
Overall good, my biggest comment is that there's a lot of lists of segments/metrics/data sources being defined throughout the tests. If possible, I think we should simplify that into one place and import throughout the tests.

tests/test_metric_libraries.py
jaredsnyder: Yeah I was wondering about that.  These tests are essentially testing the functionality of configloader, so the question is whether or not these give us anything the tests in `test_config` don't?

jaredsnyder: And if they are doing something useful we should maybe just put them in `test_config.py`

jaredsnyder: Looking it over all this is covered in `test_config.py`, I vote we delete `test_metric_librarires.py` and `test_segement_libraries.py`

tests/test_experiment.py
jaredsnyder: Added in `helpers/config_loader_lists.py`

tests/test_experiment.py
jaredsnyder: Added in helpers/config_loader_lists.py

danielkberry (approved):
LGTM

PR #234, opened 2024-05-22
(WIP) added makefile

### LINKED ISSUE #230: Evaluate using Makefile or scripts to improve usability
Many open source projects use Makefiles to standardize common tasks, we could consider doing that too to ensure that everyone installs dependencies the same way.

As part of this evaluation we should consider whether we even need a solution like this or if it would be excessive -- there might not be a real problem here.

jaredsnyder:
Fixes #230  by adding a makefile

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,10 @@
+dev_dependencies:
+	pip install -r requirements-dev.txt
+	pip install -e .
+
+test_dependencies:
+	pip install -r requirements.txt
+	pip install -e .
+
+update_deps: dev_dependencies
+	./script/update_deps


makefile
```diff
@@ -0,0 +1,10 @@
+dev_dependencies:
+	pip install -r requirements-dev.txt
+	pip install -e .
+
+test_dependencies:
+	pip install -r requirements.txt
+	pip install -e .
```
mikewilli, line 7: I think I'd suggest something like this:
```suggestion
dev_deps:
	pip install -r requirements-dev.txt

deps:
	pip install -r requirements.txt

build: deps
	pip install -e .

build_dev: dev_deps
	pip install -e .
```

mikewilli:
So this is valid and we could certainly use it, but given that the commands are so simple, it doesn't seem worth adding a new tool into the workflow at this point. Worth re-evaluating if any of the workflows get more complicated, or if we did something like incorporate docker.

PR #250, opened 2024-06-03
Improve the performance of linear model inferences

danielkberry:
This commit adds 3 performance improvements to linear model inferences that reduce the memory requirements and improve the speed: 
1. For relative inferences, instead of applying the delta method on the original data (which would require allocating several new dataframes of similar size), we create a smaller grid that approximates the empirical distribution
2. Switch to fitting the linear model by solving the normal equations. This method avoids allocating several intermediate matrices (that are the same size as the design matrix) that the built-in solvers require. 
3. Switch from Patsy to Formulaic for building the design matrix. 

With these results, Jetstream will run quickly and without running out of memory for experiments of up to 1e7 users. Experiments of larger than that (or with many branches) may still fail. As a result, I recommend increasing the memory available to each Jetstream process (by reducing the `JETSTREAM_PROCESSES` environment variable).


src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -194,23 +194,60 @@ def _extract_absolute_uplifts(
...
+        ci = results.conf_int(alpha=alpha)
+        lower, upper = ci.loc[parameter_name]
+        low_str, high_str = _stringify_alpha(alpha)
         output.loc[("abs_uplift", low_str)] = lower
         output.loc[("abs_uplift", high_str)] = upper
-
     return output
 
 
+def _create_datagrid(
```
m-d-bowerman, line 211: Have you tested to see how often conclusions from inferences using the data grid matches when the full data is used? Do you think it's worthwhile to give users the option to use the full data, rather than the approximation, when the data is small enough to easily fit in memory?

m-d-bowerman: Ah, I see the unit test comparing the results to the relative uplifts from bootstrapping. That accomplishes this, correct?

m-d-bowerman (approved):
Left some clarifying questions, for my understanding. Looks good.

src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -194,23 +194,60 @@ def _extract_absolute_uplifts(
     - output (pd.Series): the set of inferences. See `_make_joint_output`.
     """
     output = _make_joint_output(alphas, "abs_uplift")
-
     parameter_name = f"C(branch, Treatment(reference='{ref_branch}'))[T.{branch}]"
-
     output.loc[("abs_uplift", "0.5")] = results.params[parameter_name]
     output.loc[("abs_uplift", "exp")] = results.params[parameter_name]
 
+    # i = np.where(results.params.index == parameter_name)[0][0]
```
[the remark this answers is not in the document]
danielkberry: ... yes

src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -355,66 +425,112 @@ def summarize_joint(
...
+    - df (pd.DataFrame): a cleaned set of data, ready for modeling (such as the output
+    of `make_model_df`)
+    - target_col (str): the target variable for which inferences are desired.
+    - threshold_quantile (Optional[float]): the outlier threshold. See `filter_outliers`
+    - covariate_col_label (Optional[str]): the name of a covariate to include in the
+    model.
+    - copy (bool): if True (default) returns a cleaned copy of the data. If False,
+    modifies data inplace
     """
+    indexer = ~df[target_col].isna()
```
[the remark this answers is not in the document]
danielkberry: Correct, jetstream preprocessing happens before the data is provided to mozanalysis.

src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -355,66 +425,112 @@ def summarize_joint(
...
+    if covariate_col is not None:
+        indexer &= ~df[covariate_col].isna()
 
-    indexer = ~df[col_label].isna()
-    if covariate_col_label is not None:
-        indexer &= ~df[covariate_col_label].isna()
+    if copy:
```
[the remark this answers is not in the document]
danielkberry: The intention was to be defensive, in the case where a data scientist was using the functions interactively. I wanted to avoid mutating their data out from under them.

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: So comparing the relative uplifts to bootstrapping does check for this in a roundabout way, but it's probably valuable to add a more precise test. 

As far as using the option to use the full data... I think that could be a helpful extension, but I'm wondering if we can consider it out of scope for this PR and add it to the backlog?

danielkberry: 2 test cases added

m-d-bowerman: I agree re: potentially adding the feature later.

m-d-bowerman:
New tests look good, too

m-d-bowerman (approved):

PR #268, opened 2024-07-08
When covariate missing, fall back to unadjusted inferences

danielkberry:


tests/frequentist_stats/helpers.py
```diff
@@ -21,7 +21,6 @@ def __init__(self):
         self.target = "search_count"
         self.formula = "search_count ~ C(branch, Treatment(reference='treatment-a'))"
         self.results = smf.ols(self.formula, self.model_df).fit()
-        self.alphastreatment_branches = ["control", "treatment-b"]
```
mikewilli, line 24: Is this related to the PR or just some cleaning up of an unused var?

tests/frequentist_stats/test_linear_model_functions.py
```diff
@@ -593,6 +597,36 @@ def test_compare_branches_lm(interactive, deallocate):
         assert sorted(out["comparative"][branch].index) == expected_index
 
 
+def test_compare_branches_lm_fallback(caplog):
+    # tests whether the results when called with a missing covariate equal the
+    # results when called without a covariate. Also tests that a log was emitted.
+    msg = (
+        "Covariate fake_covariate not found, falling back to unadjusted inferences",
+    )[0]
```
mikewilli, line 605: Isn't this just a string?

```suggestion
    msg = "Covariate fake_covariate not found, falling back to unadjusted inferences"
```

mikewilli (approved):
Looks good to me! Just a couple minor comments.

tests/frequentist_stats/test_linear_model_functions.py
danielkberry: That's what I thought, and probably what the formatter thought too (wrapping the string in parentheses). But pytest complained about the type of the argument, saying that it got a tuple.

tests/frequentist_stats/helpers.py
danielkberry: Not related. I'm not sure where this came from, so I was trying to clean up an unused var.

tests/frequentist_stats/test_linear_model_functions.py
danielkberry: I guess not: 
Screenshot 2024-07-08 at 1 25 11 PM

danielkberry: I think the formatter must have applied the tuple wrapping when the error message was different and longer, because I [was able to](https://github.com/mozilla/mozanalysis/pull/268/commits/70b35c72218781d1f3f29d145646329fe63135bb#diff-ccf4587009a6d1005c829f95f8654a608eaf8124fbc1be967c9d731352887d55R603) just set `msg` to the correct string

mikewilli: Ah yea something like that makes sense. Thanks!

PR #269, opened 2024-07-10
Additional failure cases

danielkberry:
Also improves the type checking

### DIFF
--- a/src/mozanalysis/frequentist_stats/linear_models/functions.py
+++ b/src/mozanalysis/frequentist_stats/linear_models/functions.py
@@ -4,7 +4,6 @@
 import logging
 import re
 import warnings
-from typing import Literal, TypeAlias
 
 import numpy as np
 import pandas as pd
@@ -14,34 +13,39 @@ from pandas.api.types import is_integer_dtype
 from statsmodels.regression.linear_model import RegressionResults
 from statsmodels.stats.weightstats import DescrStatsW
 
+from mozanalysis.types import (
+    ComparativeOption,
+    CompareBranchesOutput,
+    Estimates,
+    EstimatesByBranch,
+    Uplift,
+)
+
 from .classes import MozOLS
 
 logger = logging.getLogger(__name__)
 
-##Types
-
-ComparativeOption = Literal["individual", "comparative"]
-
-Estimates: TypeAlias = pd.Series
-BranchLabel = str
-EstimatesByBranch = dict[BranchLabel, Estimates]
-
-CompareBranchesOutput = dict[ComparativeOption, EstimatesByBranch]
-
-Uplift = Literal["abs_uplift", "rel_uplift"]
 
 ## Exceptions
 
 
-class CovariateNotFound(Exception):
+class LinearModelException(Exception):
     pass
 
 
-class UnableToAnalyze(Exception):
+class CovariateNotFound(LinearModelException):
     pass
 
 
-class FailedToFitModel(Exception):
+class UnableToAnalyze(LinearModelException):
+    pass
+
+
+class FailedToFitModel(LinearModelException):
+    pass
+
+
+class MissingModelParameter(LinearModelException):
     pass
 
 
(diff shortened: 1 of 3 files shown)


src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -17,6 +18,35 @@
...
+class CovariateNotFound(Exception):
+    pass
+
+
+class UnableToAnalyze(Exception):
+    pass
+
+
+class FailedToFitModel(Exception):
+    pass
```
mikewilli, line 45: You could have these inherit from a `CovariateException` or something, which would inherit from `Exception`, if you want to group them in some way. Not sure if that accomplishes anything meaningful though if they are all just `Exception` by a different name.

src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -17,6 +18,35 @@
 
 logger = logging.getLogger(__name__)
 
+##Types
+
+ComparativeOption = Literal["individual", "comparative"]
```
mikewilli, line 23: Typically I'd choose to use an Enum here (something like below), is Literal a better option in this case?

```py
class ComparativeOption(str, Enum):
    INDIVIDUAL = "individual"
    COMPARATIVE = "comparative"
````

File: src/mozanalysis/frequentist_stats/linear_models/functions.py [hunk 3]
```diff
@@ -334,27 +369,58 @@ def fit_model(
 
     Returns:
     - results (RegressionResults): the fitted model results object.
+
+    Raises:
+    - `FailedToFitModel` if unable to fit the model.
+    - `Exception` if certain unexpected conditions arise (e.g., one of the treatment)
+    branches, which was expected to have a model parameter, does not get one.
```
mikewilli, line 376: Typo, but also -- should we have a more specific Exception for this case? I think it is usually preferred to raise a more descriptive exception type whenever possible (this makes it easier for exception handling because it allows the calling code to catch a specific exception instead of having to catch `Exception` which would be anything). It doesn't necessarily have to be a custom Exception type that we create, but that just depends on what's happening.

```suggestion
    - `Exception` if certain unexpected conditions arise (e.g., one of the treatment
    branches, which was expected to have a model parameter, does not get one).
```

src/mozanalysis/frequentist_stats/linear_models/functions.py
```diff
@@ -589,6 +701,10 @@ def compare_branches_lm(
     and `summarize_joint` for more information on the output structure.
 
     """
+
+    if alphas is None:
+        alphas = [0.01, 0.05]
+
```
mikewilli, line 707: I'd suggest we just make this the default for the parameter in the function signature.

mikewilli (approved):
r+wc -- Nothing major, though I'd suggest that we try to avoid raising a regular `Exception`. Other ones are more thoughts than suggestions, up to you.

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: tbh, I feel like an Enum is probably better. That way Jetstream can even import and use it.

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: Ignored to avoid mutable data structures as default parameter values

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: Done

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: Done

src/mozanalysis/frequentist_stats/linear_models/functions.py
danielkberry: Done

PR #270, opened 2024-07-11
chore: more precise errors in tests; README typos

mikewilli:
Was playing around with the tests for something else and noticed that we were checking for `Exception` when we could be more specific. Also saw some extremely minor typos in the README that I fixed for consistency.


danielkberry (approved):
These are great improvements!

PR #272, opened 2024-07-29
Update config.py

danielkberry:


src/mozanalysis/config.py
```diff
@@ -51,7 +77,12 @@ def get_metric(self, metric_slug: str, app_name: str):
 
         metric_definition = self.configs.get_metric_definition(metric_slug, app_name)
         if metric_definition is None:
-            raise Exception(f"Could not find definition for metric {metric_slug}")
+            if self.check_configs_for_app(app_name):
+                raise MetricNotFound(
+                    f"Could not find definition for metric {metric_slug}"
+                )
+            else:
+                raise ApplicationNotFound(f"Could not find application {app_name}")
```
mikewilli, line 85: Would it still be helpful to include the other info here?
```suggestion
                raise MetricNotFound(
                    f"Could not find definition for metric {metric_slug} in application {app_name}"
                )
            else:
                raise ApplicationNotFound(f"Could not find application {app_name}, so metric {metric_slug} could not be resolved")
```

mikewilli: Same comment for the other changes if you like this idea.

mikewilli (approved):
Suggested including some extra info in the error messages, but looks good!

jaredsnyder (approved):
looks great!

PR #273, opened 2024-07-31
DS-3791: Groupid aggregation

danielkberry:
Fulfills [DS-3791: Modify mozanalysis to support groupID level aggregation](https://mozilla-hub.atlassian.net/browse/DS-3791). 

Changes: 
* Adds an `AnalysisUnit` Enum to denote the [experimental unit](https://www.britannica.com/science/experimental-unit) of the experiment. Possible values are `CLIENT` (mapping to `client_id`) or `GROUP` (mapping to `profile_group_id`)
* Adds an `analysis_unit: AnalysisUnit` attribute to the `Experiment` class to denote the unit for the experiment. The experimental unit is now used to define the join key in the query builders: 
  * `Experiment.build_enrollments_query` 
  * `Experiment._build_enrollments_query` now gates `Experiment._build_enrollments_query_<subtype>` methods  by checking for valid values of `experimental_unit` for the given `enrollment_query_type`. 
  * `Experiment.build_exposure_query` (`Experiment._build_exposures_query` performs similar gating as `Experiment._build_enrollments_query`)
  * `Experiment.build_metrics_query`
* Adds a `analysis_unit: AnalysisUnit` attribute to the `TimeSeriesResult` class (used when joining multiple windows together). 
* `DataSource`: 
  * `DataSource`s can support both analysis units simultaneously (e.g., data may be keyed by both `client_id` and `profile_group_id`). Thus, the analysis unit should be passed in at runtime, not at instantiation time. 
  * Adds a `group_id_column: str` attribute to `DataSource` to mark the name of the group_id column. 
  * Adds a `analysis_unit: AnalysisUnit` parameter to `DataSource.build_query` to define the level to which data should be aggregated when building a particular query. 
* Adds an `IncompatibleAnalysisUnit` error type. 
* Some minor changes were made to clean up typing (`mypy` now runs without complaint for `experiment.py`)
* Unit tests for enrollments and metrics query building for both client and group levels 
* Downsampling not allowed for group-level experiments until decision made on group-level sample_ids

Possible Extensions: 
  * We could consider moving `EnrollmentsQueryType` from a runtime parameter passed to the query methods to being an attribute of the `Experiment` object. 
    * As best I can tell, the query type should be static across the lifetime of the `Experiment` object, so we could make it an attribute. This would allow us to move the above validity check to `Experiment` instantiation time, so that the check happens as early as possible and isn't necessary to repeat in `Experiment._build_enrollments_query` and `Experiment._build_exposure_query`. 
    * Currently, `CLIENT` experimental units are supported for all `EnrollmentsQueryType`s, while `GROUP` is only supported for `EnrollmentsQueryType.NORMANDY`. 
  * Should `AnalysisUnit` live in metric-config-parser instead?
  
Considered Alternatives: 
* I considered simply modifying the initial `SELECT` portions of the query to rename `group_id` to `client_id` (i.e., `SELECT profile_group_id AS client_id`), which would have simplified the necessary changes. This option was rejected due to risk of subtle errors (e.g., one step is missed and group-level data was joined to client-level data). 

Notes: 
* `profile_group_id` is not yet on the `telemetry.events` view, but is [scheduled to be](https://docs.google.com/spreadsheets/d/1QGKvd7NkO0Icod8-yQEvOWxqVQ7C0Y9-x_PE9rbssug/edit?gid=1915465914#gid=1915465914).


src/mozanalysis/types.py
```diff
@@ -19,3 +19,8 @@ class Uplift(str, Enum):
 EstimatesByBranch = dict[BranchLabel, Estimates]
 
 CompareBranchesOutput = dict[ComparativeOption, EstimatesByBranch]
+
+
+class ExperimentalUnit(str, Enum):
+    CLIENT = "client_id"
+    GROUP = "profile_group_id"
```
jmsilverman, line 26: Here and in the queries elsewhere, are we sure the field is called `profile_group_id` and not `group_id` or am I misreading something?

src/mozanalysis/experiment.py
```diff
@@ -654,37 +696,67 @@ def _build_exposure_query(
...
+                `moz-fx-data-shared-prod.telemetry.events` e
+            WHERE
+                e.event_category = 'normandy'
+                AND e.event_method = 'enroll'
+                AND e.submission_date
+                    BETWEEN '{time_limits.first_enrollment_date}' AND '{time_limits.last_enrollment_date}'
+                AND e.event_string_value = '{self.experiment_slug}'
+                AND e.sample_id < {sample_size}
+            GROUP BY e.profile_group_id, branch
+                """  # noqa:E501
```
jmsilverman, line 757: Is it worth using a string var for client_id vs profile_group_id in an if statement outside the query so we don't have to have 2 copies of effectively the same query?

src/mozanalysis/experiment.py
```diff
@@ -798,32 +873,62 @@ def _build_enrollments_query_cirrus(
...
+                    AND submission_date
+                        BETWEEN '{time_limits.first_enrollment_date}' AND '{time_limits.last_enrollment_date}'
+                    AND event_string_value = '{self.experiment_slug}'
+            ) e
+            ON re.profile_group_id = e.profile_group_id AND
+                re.branch = e.branch AND
+                e.submission_date >= re.enrollment_date
+            GROUP BY e.profile_group_id, e.branch
+                """  # noqa: E501
+        else:
```
jmsilverman, line 930: Same comment as above around not having 2 copies of nearly the same query.

src/mozanalysis/experiment.py
```diff
@@ -654,37 +696,67 @@ def _build_exposure_query(
...
+                COUNT(e.submission_date) AS num_enrollment_events
+            FROM
+                `moz-fx-data-shared-prod.telemetry.events` e
+            WHERE
+                e.event_category = 'normandy'
+                AND e.event_method = 'enroll'
+                AND e.submission_date
+                    BETWEEN '{time_limits.first_enrollment_date}' AND '{time_limits.last_enrollment_date}'
+                AND e.event_string_value = '{self.experiment_slug}'
+                AND e.sample_id < {sample_size}
```
jmsilverman, line 734: I believe sample_id is currently based on client_id, do we know if that will continue into the future? Will there be a group_sample_id or something that will be based on the profile_group_id or group_id column?

danielkberry: Ohh yikes, good question... group_ids are not guaranteed to all fall into the same sample_ids... that's a huge footgun

src/mozanalysis/experiment.py
danielkberry: Initially had them separate because they could vary... they ended up initially varying much less than than expected. Why don't we prohibit downsampling for group level experiments until the group level sampling is resolved

src/mozanalysis/types.py
danielkberry: You are not, apparently that's how it was implemented ¯\_(ツ)_/¯

jmsilverman: Yup, confirmed elsewhere, feel free to ignore this comment.

src/mozanalysis/experiment.py
jmsilverman: Even if we did that, which is reasonable, we could still use string formatting+variables to do it I think...though that does make the queries more distinct. I just personally never like seeing more than a few line SQL queries like this copied & pasted, especially when they already use a few string format vars, but I leave it up to you.

src/mozanalysis/experiment.py
jmsilverman: Confirming this is an open question with a Jira ticking coming to track the work around it and ongoing discussions are happening among DS and DE.

src/mozanalysis/experiment.py
danielkberry: Changed.

src/mozanalysis/experiment.py
danielkberry: Done, one version of the query now

src/mozanalysis/experiment.py
```diff
@@ -1141,6 +1214,8 @@ def for_ts(
             ]
         )
 
+        analysis_windows = cast(tuple[AnalysisWindow], analysis_windows)
```
mikewilli, line 1217: This feels a bit inane for ruff to complain about (assuming that's what happened here). Maybe we could define the type with the definition (`analysis_windows: tuple[AnalysisWindow] = ...`) to make it more concise? If it wasn't ruff or you prefer this then no worries!

src/mozanalysis/experiment.py
```diff
@@ -1354,11 +1430,11 @@ def _build_analysis_window_subset_query(
         in "the standard format" for a single analysis window.
         """
         return f"""
-            SELECT * EXCEPT (client_id, analysis_window_start, analysis_window_end)
+            SELECT * EXCEPT ({self.experimental_unit.value}, analysis_window_start, analysis_window_end)
             FROM {self.fully_qualified_table_name}
             WHERE analysis_window_start = {analysis_window.start}
             AND analysis_window_end = {analysis_window.end}
-        """
+        """  # noqa: E501
```
mikewilli, line 1437: Could you just split the `EXCEPT ( ... )` into multiple lines? I usually try to avoid `noqa` whenever possible, though I understand it can be a bit annoying so not a big deal if splitting the string introduces other issues.

src/mozanalysis/metrics.py
```diff
@@ -141,16 +146,22 @@ def build_query(
...
-        return """
-        SELECT
-            e.client_id,
+        if experimental_unit == ExperimentalUnit.CLIENT:
+            ds_id = self.client_id_column or "client_id"
+        elif experimental_unit == ExperimentalUnit.GROUP:
+            ds_id = self.group_id_column or "profile_group_id"
+        else:
+            assert_never(experimental_unit)
+        return """SELECT
```
mikewilli, line 163:
```suggestion
            assert_never(experimental_unit)

        return """SELECT
```

src/mozanalysis/metrics.py
```diff
@@ -141,16 +146,22 @@ def build_query(
...
         This query does not define ``enrollments`` but otherwise could
         be executed to query all metrics from this data source.
         """
-        return """
-        SELECT
-            e.client_id,
+        if experimental_unit == ExperimentalUnit.CLIENT:
+            ds_id = self.client_id_column or "client_id"
+        elif experimental_unit == ExperimentalUnit.GROUP:
+            ds_id = self.group_id_column or "profile_group_id"
```
mikewilli, line 160: It shouldn't be necessary to specify `or "client_id"`, right? The default is set on the class attr, so unless the config specifically defines it as an invalid falsy value, I don't know what this would do. I guess probably safer to just leave it in since this was the construction already... What do you think?

src/mozanalysis/metrics.py
```diff
@@ -206,6 +221,11 @@ def build_query_targets(
         This query does not define ``targets`` but otherwise could
         be executed to query all metrics from this data source.
         """
+        if experimental_unit != ExperimentalUnit.CLIENT:
+            raise ValueError(
+                "`build_query_targets` currently supports client_id analysis"
+            )
```
mikewilli, line 227:
```suggestion
            raise IncompatibleExperimentalUnit(
                "`build_query_targets` currently only supports client_id analysis"
            )
```

src/mozanalysis/types.py
mikewilli, line 26: What do you think about calling this `PROFILE_GROUP`?

src/mozanalysis/metrics.py
```diff
@@ -69,6 +73,7 @@ class DataSource:
     submission_date_column = attr.ib(default="submission_date", type=str)
     default_dataset = attr.ib(default=None, type=str | None)
     app_name = attr.ib(default=None, type=str | None)
+    group_id_column = attr.ib(default="profile_group_id", type=str)
```
mikewilli, line 76: Should this and `client_id_column` above use `ExperimentalUnit.GROUP.value`? And elsewhere that we have the strings spelled out (like in `build_query`) we could use these values as well.

mikewilli (commented):
Haven't had a chance to review the tests but thought it might be helpful to get these comments to you today. I don't think I saw any major issues though, nice work!

src/mozanalysis/types.py
```diff
@@ -19,3 +19,8 @@ class Uplift(str, Enum):
 EstimatesByBranch = dict[BranchLabel, Estimates]
 
 CompareBranchesOutput = dict[ComparativeOption, EstimatesByBranch]
+
+
+class ExperimentalUnit(str, Enum):
```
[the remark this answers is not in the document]
danielkberry: "Experimental unit" means both "randomization unit" and "analysis unit". Right now, our randomization unit (per-install, sub-profile level) differs from our analysis unit (profile level, keyed by client_id). One of the big advantages of switching to group ID is that that synchronizes the two units, which has some nice benefits (or rather, removes some weird downsides). 

Now, given what I said above, I suppose we _should_ call this the `AnalysisUnit`, because for legacy experiments this is only the analysis unit, not exactly the randomization unit. I actually had it that way, but [switched](https://github.com/mozilla/mozanalysis/pull/273/commits/380b00d6de0a2028f3421a05a8dfac59542005f1).  

Thoughts?

src/mozanalysis/metrics.py
danielkberry: I'm inclined to agree. As you said, I did it this way b/c that's how it already was done, but I don't _think_ `client_id_column` should ever be null...

src/mozanalysis/types.py
danielkberry: I think that would add some clarity, especially since that's what the ID was named.

src/mozanalysis/metrics.py
danielkberry: Good callout.

src/mozanalysis/types.py
danielkberry: Done

src/mozanalysis/metrics.py
danielkberry: Actually, the `type=str` attrs parameter doesn't _validate_ the type at runtime. See below example for a metric pulled from `ConfigLoader` which has a `None` where we would expect a string
Screenshot 2024-08-02 at 8 56 37 AM

danielkberry: I propose we move these `or default_value` guards to the attrs field definition, something like: 

```python
client_id_column = attr.ib(
    default=ExperimentalUnit.CLIENT.value,
    type=str,
    validator=[attr.validators.instance_of(str), attr.validators.min_len(1)],
    converter=lambda x: "client_id" if x is None else x
)
```

that way, anywhere inside the class, `self.client_id_column` is guaranteed to be a valid string

danielkberry: ok, I made this change. the `or "client_id"` check can be removed, now guaranteed to be a string at runtime. Also made the change for group_id_column and submission_date_column

src/mozanalysis/metrics.py
danielkberry: Fixed

src/mozanalysis/types.py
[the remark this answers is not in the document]
mikewilli: Thanks for the clarification! I think in that case it makes sense to call it `AnalysisUnit`.

src/mozanalysis/experiment.py
```diff
@@ -629,12 +657,18 @@ def _build_enrollments_query(
                 raise ValueError(
                     "App ID must be defined for building Cirrus enrollments query"
                 )
+            if not self.experimental_unit == ExperimentalUnit.CLIENT:
+                raise IncompatibleExperimentalUnit(
+                    "Cirrus enrollments currently only support client_id analysis units"
+                )
```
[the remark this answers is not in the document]
mikewilli: So should we just ignore the `experimental_unit` here? Or is this intended to prevent an invalid configuration where a Cirrus experiment has an `experimental_unit` when it shouldn't?

src/mozanalysis/experiment.py
mikewilli: Gotcha, works for me. Does being explicit about the type during the definition as I suggested above not work? That's the only idea I have, so if it doesn't work maybe we just stick with the cast.

As a side note, I wonder if we were too greedy in switching to ruff only. I'd say it's worth putting mypy back in as a requirement (for both mozanalysis and jetstream) -- the ruff FAQ suggests running them together.

mikewilli: Created tickets to add mypy back in.

src/mozanalysis/metrics.py
mikewilli: Awesome, thanks! And I agree on the metric-config-parser suggestion as well.

src/mozanalysis/metrics.py
```diff
@@ -320,6 +376,30 @@ def get_sanity_metrics(self, experiment_slug: str) -> list[Metric]:
         else:
             raise ValueError
 
+    @classmethod
+    def from_mcp_data_source(
+        cls,
+        parser_data_source: ParserDataSource,
+        app_name: str | None = None,
+        group_id_column: str | None = ExperimentalUnit.PROFILE_GROUP.value,
+    ) -> DataSource:
```
mikewilli, line 385: I was going to say that I don't love the `mcp_data_source` naming here, but realized I don't have a good solution. So instead I'll just note that I wish we didn't have the same (or nearly the same) objects defined all over the place.  I guess another symptom of the many-repo issue we've discussed at length..

danielkberry: Yeah, I agree with you. Some ideas I've got: 
- rename mcp objects into "definition"s, then mozanalysis objects can have a "from_definition" method
- instead of mozanalysis's DataSource largely duplicating the parameters, maybe it should wrap the mcp object and only take the extra parameters it needs which, afaict, is just app_id (which isn't present in the data source definition and is provided by jetstream at runtime)

src/mozanalysis/types.py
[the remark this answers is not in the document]
danielkberry: Awesome, I'll update.

src/mozanalysis/metrics.py
mikewilli: I think I like the second one, and maybe we call the mozanalysis version `ExperimentDataSource` or something? What do you think?

Either way I'm OK with leaving as-is in this PR, better to separate out that complexity.

danielkberry: +1. I also like the 2nd one better, and agree with doing it later, not in this PR

src/mozanalysis/experiment.py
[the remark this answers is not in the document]
danielkberry: My thinking was to be explicit and fail if the experimental unit was used anywhere other than desktop experiments (normandy enrollments). Alternatively, we could ignore it (possibly logging a warning). My concern is that misconfiguration here could be subtle (there wouldn't be any way to directly tell from the results whether the configuration was applied), so if someone tries to do the wrong thing, they might think that it worked, which would be confusing.

src/mozanalysis/metrics.py
danielkberry: Great, [issue filed](https://github.com/mozilla/metric-hub/issues/545)

src/mozanalysis/metrics.py
danielkberry: Issue filed: https://github.com/mozilla/mozanalysis/issues/293

src/mozanalysis/experiment.py
danielkberry: Done

src/mozanalysis/experiment.py
danielkberry: I changed the type signature to a variable length tuple, now mypy doesn't complain about the type and we don't need a cast

src/mozanalysis/experiment.py
```diff
@@ -129,11 +137,16 @@ class Experiment:
...
-    experiment_slug = attr.ib()
+    experiment_slug = attr.ib(type=str, validator=attr.validators.instance_of(str))
     start_date = attr.ib()
     num_dates_enrollment = attr.ib(default=None)
     app_id = attr.ib(default=None)
     app_name = attr.ib(default=None)
+    analysis_unit = attr.ib(
+        type=AnalysisUnit,
+        default=AnalysisUnit.CLIENT,
+        converter=lambda s: AnalysisUnit(s),  # allows callers to pass a string
```
mikewilli, line 148: Is it desirable to allow calls to pass a string?

src/mozanalysis/metrics.py
```diff
@@ -140,17 +186,24 @@ def build_query(
         time_limits: TimeLimits,
         experiment_slug: str,
         from_expr_dataset: str | None = None,
-        analysis_basis: str = AnalysisBasis.ENROLLMENTS,
+        analysis_basis: AnalysisBasis = AnalysisBasis.ENROLLMENTS,
+        experimental_unit: AnalysisUnit = AnalysisUnit.CLIENT,
```
mikewilli, line 190: Should we use `analysis_unit` here, or is `experimental_unit` more correct in this case?

mikewilli (approved):
Heck yea, great work and love all the tests!

A couple outstanding questions to address (including leftover from previous comments), but they're minor enough that I'm happy to approve in the meantime.

.circleci/config.yml
```diff
@@ -44,12 +41,12 @@ jobs:
...
-    - run:
-        name: Run linting
-        command: |
-          pip install ruff
-          ruff check src/ tests/
+      - checkout
+      - run:
+          name: Run linting
+          command: |
+            pip install -r requirements.txt
```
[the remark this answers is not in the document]
danielkberry: Before the change it was using ruff version 0.6.1, while our deps have that pinned to 0.5.7. That was causing CI to fail, but not in the local environment. I talked to @jaredsnyder and we decided the simplest thing to do would be to just install all deps here. We considered attempting to install just the correct ruff version (grep-ing the requirements file?) but that seemed fragile. But no, we just need ruff installed.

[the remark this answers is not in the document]
mikewilli: Ah yea that makes sense, thanks! I agree with your decision.

src/mozanalysis/metrics.py
danielkberry: No, I think that's old and needs to be updated. `analysis_unit` should be right.

src/mozanalysis/experiment.py
danielkberry: I thought it was back when `AnalysisUnit` lived in mozanalysis, but now that the object is defined externally, in metric-hub, I think it's probably safe to force callers to use the enum

src/mozanalysis/experiment.py
```diff
@@ -414,7 +426,7 @@ def get_time_series_data(
             fully_qualified_table_name=bq_context.fully_qualify_table_name(
                 full_res_table_name
             ),
-            analysis_windows=time_limits.analysis_windows,
+            analysis_windows=list(time_limits.analysis_windows),
```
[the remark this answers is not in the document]
danielkberry: I'm not sure, but let's revisit this later

src/mozanalysis/experiment.py
danielkberry: Filed [a ticket](https://mozilla-hub.atlassian.net/jira/software/c/projects/DS/boards/1947?selectedIssue=DS-3857) to follow up on this, not going to block this PR on it.

PR #294, opened 2024-08-12
update deps

jaredsnyder:
dependabot is angry


rebecca-burwei (approved):
pytest passed locally

PR #295, opened 2024-08-22
Remove downsampling restriction for group level experiments

danielkberry:
As per [this comment](https://mozilla-hub.atlassian.net/browse/DS-3857?focusedCommentId=924733), the decision which enables sampling by profile group ID has been approved. This PR removes the restriction preventing downsampling for group level experiments.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -716,10 +716,6 @@ class Experiment:
         sample_size: int = 100,
     ) -> str:
         """Return SQL to query enrollments for a normandy experiment"""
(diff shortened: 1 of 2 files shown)


mikewilli:
I can't read the decision doc (requested access), but `sample_id` will continue to work across both `client_id` and `profile_group_id` simultaneously?

danielkberry:
Right. The solution will be the set the group ID to be the client ID for existing clients. That way those profiles, and the data from all other profiles, is guaranteed to fall into the same sample id

mikewilli (approved):

PR #320, opened 2024-09-12
Update deps

jaredsnyder:
satisfy dependabot, ruff update required some changes but they should all be cosmetic (just ran `ruff check --fix`


tests/frequentist_stats/test_sample_size_calc.py
```diff
@@ -19,7 +20,7 @@
 uri_count = ConfigLoader.get_metric("uri_count", "firefox_desktop")
 
 
-@pytest.fixture()
```
danielkberry, line 22: Was this an automated change?

danielkberry (approved):
Overall, looks pretty good. There was one question where it looks like a method call was removed, was that intended and is that ok?

tests/frequentist_stats/test_sample_size_calc.py
jaredsnyder: Yeah, I'm not sure how this would change things but the tests that use the fixture still work so 🤷

jaredsnyder: Seems like it's normal: https://docs.pytest.org/en/6.2.x/fixture.html

PR #321, opened 2024-09-13
add analysis unit support for segments

mikewilli:


src/mozanalysis/segments.py
```diff
@@ -45,10 +49,11 @@ class SegmentDataSource:
     _from_expr = attr.ib(validator=attr.validators.instance_of(str))
     window_start = attr.ib(default=0, type=int)
     window_end = attr.ib(default=0, type=int)
-    client_id_column = attr.ib(default="client_id", type=str)
+    client_id_column = attr.ib(default=AnalysisUnit.CLIENT.value, type=str)
     submission_date_column = attr.ib(default="submission_date", type=str)
     default_dataset = attr.ib(default=None, type=str | None)
     app_name = attr.ib(default=None, type=str | None)
+    group_id_column = attr.ib(default=AnalysisUnit.PROFILE_GROUP.value, type=str)
```
scholtzan, line 56: Do we need to update metric-config-parser as well? https://github.com/mozilla/metric-hub/blob/6ddcc3220488d06df5e8b8b35125bde369f4a90c/lib/metric-config-parser/metric_config_parser/data_source.py#L105

scholtzan (approved):
Does anything need to be updated for the population sizing as well?

This looks good to me. Also as the default is analysis based on `client_id` it shouldn't break anything that is currently using it.

src/mozanalysis/segments.py
mikewilli: Yep, I thought @danielkberry and I had discussed this when I did the `analysis_unit` PR, but can't find the conversation now. Unless there's something I'm forgetting there I don't see why we wouldn't want it to be configurable explicitly.

mikewilli:
Yes, the sizing stuff will need to be updated as well in order to work with groups. I'm not planning on tackling that immediately but makes sense to just follow up the jetstream-related pieces with that while it's on the mind.

PR #324, opened 2024-09-18
use analysis_id instead of client/profile_group for consistency

mikewilli:
Before this change, the analysis tables would have an inconsistent identifier column called either `client_id` or `profile_group_id`, depending on the experiment. This change renames that column to `analysis_id` for all experiments.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -507,10 +507,10 @@ class Experiment:
 
             SELECT
                 se.*,
-                e.* EXCEPT ({self.analysis_unit.value}, branch)
+                e.* EXCEPT (analysis_id, branch)
             FROM segmented_enrollments se
             LEFT JOIN exposures e
-            USING ({self.analysis_unit.value}, branch)
+            USING (analysis_id, branch)
         """
 
     def build_metrics_query(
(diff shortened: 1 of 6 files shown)


scholtzan (approved):

src/mozanalysis/experiment.py
```diff
@@ -771,7 +774,7 @@ def _build_enrollments_query_fenix_baseline(
                 '{experiment_slug}'
             ).branch IS NOT NULL
             AND b.sample_id < {sample_size}
-        GROUP BY client_id, branch
+        GROUP BY b.client_info.client_id, branch
```
danielkberry, line 777: Do we run `fenix_baseline` experiments any more? Was this change related to the PR or is this fixing an old issue?

danielkberry (approved):
LGTM! Thanks for tackling this

src/mozanalysis/experiment.py
mikewilli: Not sure on the first part, but I was just updating everything where I either saw an issue or wanted to make things consistent. I think this one in particular was an issue because I renamed the alias client_id to analysis_id and so I could no longer group on client_id. However, I'm pretty sure the way SQL is evaluated means you shouldn't be able to use aliases in GROUP BY anyway (bigquery might do some stuff here to make it work), so instead of changing to analysis_id I just went with the full original field.

PR #325, opened 2024-09-19
add warning if wrong analysis_unit in custom query

mikewilli:
Updated some more comments about custom queries to reflect the change in queries from client_id to analysis_id.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -447,6 +447,7 @@ class Experiment:
         exposure_signal: ExposureSignal | None = None,
         segment_list=None,
         sample_size: int = 100,
+        suppress_custom_query_validation: bool = False,
     ) -> str:
         """Return a SQL query for querying enrollment and exposure data.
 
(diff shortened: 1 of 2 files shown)


danielkberry (approved):
This looks good, but how are the warnings surfaced to the user? Will they just be emitted during the CI but the CI will otherwise pass the configuration?

mikewilli:
Yea, I was a little concerned about raising an exception in case there was some valid reason to have the alternative Analysis Unit string value in the query somewhere. Maybe that's not a concern and we could raise an exception instead?

Or perhaps we could raise an exception here but handle it in Jetstream so it doesn't crash but does produce a log in the Errors dashboard?

mikewilli:
OK, based on discussions with @danielkberry, I changed the warning log to raise a `ValueError` instead. I also added a flag that will allow the caller to bypass this check (and instead log a warning if the check fails), in case there is a situation where the check is erroneous or the caller is intentionally putting the wrong analysis unit in the custom query.

PR #351, opened 2024-10-09
chore: update docs-deploy to supported dep versions

mikewilli:
Updates node base image and gh-pages dep to latest because
- node version 8.10.0 is 7 years old
- gh-pages version 2.0.1 is 6 years old


jaredsnyder (approved):
lfgggggggggggggg

mikewilli:
Seems like `gh-pages` >=3 causes an error with our current command. Not going to spend any time looking into it further right now, so for now I'll just get the base image updated and leave it at that.

PR #398, opened 2024-12-23
upgrade deps

jaredsnyder:
Upgrade dependencies based on dependabot requirements


jaredsnyder:
Validated the envs can install locally, and that all the deps except marginal effects) got updated

m-d-bowerman (approved):

PR #431, opened 2025-02-11
Query events_stream table for Glean experiment enrollment

curtismorales:
Changes the experiment enrollment query for Glean experiments to query `events_stream` instead of `events`. This should be a more efficient query.

### DIFF
--- a/src/mozanalysis/experiment.py
+++ b/src/mozanalysis/experiment.py
@@ -712,7 +712,7 @@ class Experiment:
                 raise IncompatibleAnalysisUnit(
                     "Glean enrollments currently only support client_id analysis units"
                 )
-            return self._build_enrollments_query_glean_event(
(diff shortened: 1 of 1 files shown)


mikewilli (changes requested):
Thanks! One thing I forgot earlier, can you bump the version too in pyproject? CalVer so should be 2025.2.1.

And I'll follow up in Jetstream by updating the integration tests for the new events_stream table alongside the existing tests for events.

Fixes https://github.com/mozilla/jetstream/issues/2325

mikewilli:
Oh right, I forgot mozanalysis just uses the git tag, my bad!

mikewilli (approved):
Looks good to me! Sorry for the above confusion -- we can tag the release once it's merged and that'll trigger a deployment.

PR #457, opened 2025-03-14
fix: ruff lint rule TCH name change

mikewilli:
* ruff renamed the `TCH` rule to `TC`
* sort the ruff lint rules
* update ruff to latest (0.11.0)


src/mozanalysis/experiment.py
```diff
@@ -1060,7 +1060,7 @@ def _build_metrics_query_bits(
                 metrics.append(metric)
 
         ds_metrics = partition_metrics_by_data_source(metrics)
-        ds_metrics = cast(dict[DataSource, list[Metric]], ds_metrics)
+        ds_metrics = cast("dict[DataSource, list[Metric]]", ds_metrics)
```
jaredsnyder, line 1063: What does this change mean?

jaredsnyder (approved):
Changes make sense and only affect the linting.  Linting is running fine in the CI so think we're good here!

src/mozanalysis/experiment.py
mikewilli: Good question. ruff thinks we should quote the types when calling `cast`. `cast` doesn't actually do anything at runtime and is merely for type hints, so practically it doesn't do anything.

As for _why_ ruff thinks we should do this, it appears to be a [style/consistency thing](https://docs.astral.sh/ruff/rules/runtime-cast-value/). Basically because `cast` calls _can_ include forward references (which must be quoted), ruff thinks we should quote every type passed to `cast`.

PR #460, opened 2025-03-14
chore: fix warnings raised by tests

mikewilli:
Noticed there were a handful of warnings coming out of running pytest due to some deprecated calls, or in some cases expected behavior that we weren't accounting for in the tests (e.g., UserWarnings that we expect to raise, or side effects of our use of 0'd out dataframes for tests).


danielkberry (approved):
Wonderful quality of life improvements! Thanks Mike!

PR #485, opened 2025-04-02
feat(nimbus): Update to nimbus user id

yashikakhurana:
Because we have updated from user_id to nimbus_user_id https://github.com/mozilla/experimenter/pull/12017/files ,  so updating to reflect the same here


mikewilli (approved):
LGTM thanks!

PR #507, opened 2025-05-05
Add new experiments_column_type to support events_stream tables

relud:
fixes https://github.com/mozilla/jetstream/issues/2174

I'm unclear if there's more plumbing this needs.

I went for a condition that's going to match the other type's semantics of checking both experiment and branch, but I don't understand enough to tell if that's actually necessary or preferrable.

### DIFF
--- a/src/mozanalysis/metrics.py
+++ b/src/mozanalysis/metrics.py
@@ -116,7 +116,7 @@ class DataSource:
         converter=group_id_column_converter,
     )
 
-    EXPERIMENT_COLUMN_TYPES = (None, "simple", "native", "glean")
+    EXPERIMENT_COLUMN_TYPES = (None, "simple", "native", "glean", "events_stream")
 
     @experiments_column_type.validator
     def _check_experiments_column_type(self, attribute, value):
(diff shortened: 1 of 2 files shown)


src/mozanalysis/metrics.py
```diff
@@ -413,6 +423,32 @@ def get_sanity_metrics(self, experiment_slug: str) -> list[Metric]:
...
+            return [
+                Metric(
+                    name=self.name + "_has_contradictory_branch",
+                    data_source=self,
+                    select_expr=agg_any(
+                        """IF(
+                JSON_VALUE(ds.event_extra, '$.experiment') = '{experiment_slug}',
+                JSON_VALUE(ds.event_extra, '$.branch'),
+                NULL
+            ) != e.branch """
```
mikewilli, line 436: Yea I think you're right that this matches the other checks a little more closely than just checking the branch alone, and either way it's no harm to check the experiment + branch.

tests/test_metrics.py
```diff
@@ -7,7 +7,9 @@
 from mozanalysis.metrics import AnalysisBasis, DataSource, Metric
 
 
-@pytest.mark.parametrize("experiments_column_type", [None, "simple", "native", "glean"])
+@pytest.mark.parametrize(
+    "experiments_column_type", [None, "simple", "native", "glean", "events_stream"]
+)
 def test_datasource_constructor_succeeds(experiments_column_type):
     DataSource(
         name="foo",
```
mikewilli, line 15: It seems like there should be more tests than this (e.g., for the two functions you modified here). I know we have more comprehensive testing in jetstream so maybe not a big deal, but if you've got some time and inclination now want to tackle that here? Otherwise can you just file a ticket (in mozanalysis issues is good) to add unit tests for those two functions?

mikewilli (approved):
Awesome thanks!

mikewilli:
@relud Want to land this? Or are you working on the tests?

PR #509, opened 2025-05-29
fix: group_id_column carried over from mcp DS

mikewilli:
We added `group_id_column` here before it was supported in metric-config-parser, but never updated once mcp added support so the config is allowed but not actually used. This rectifies that situation and so it should be used after this change (and after jetstream updates to use it).


danielkberry (approved):
LGTM

PR #510, opened 2025-05-29
fix: sanity metric query strings should be f-strings

relud:
I don't know if this is correct, but when I added `experiments_column_type == "events_stream"`, I saw that the first metric doesn't use f-string and the second metric does, even though both reference experiment_slug.


mikewilli (approved):
Hm, strange... looking at the queries being built, it appears that both work. That said, not going to argue against consistency.

PR #516, opened 2025-08-29
Update marginaleffects version

danielkberry:
We were blocked on a bugfix landing upstream, that's in now so we should be able to update to the latest version


mikewilli (approved):
LGTM, no changes in how we use it?

PR #518, opened 2025-10-20
[DENG-8817] Migrate mozanalysis CircleCI to Github Actions

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-8817

Testing out https://github.com/mozilla/circleci-to-gha to do this migration

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,11 +40,16 @@ jobs:
         run: |
           pip install tox
           tox -e py310
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v4
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          flags: ${{ github.event.repository.name }}
+      - name: Submit code coverage data
+        run: |
+          curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
+          curl -Os https://uploader.codecov.io/latest/linux/codecov
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM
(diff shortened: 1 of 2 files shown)


scholtzan:
I'll disable CircleCI once this has been approved and is ready to get merged

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+          curl -Os https://uploader.codecov.io/latest/linux/codecov.SHA256SUM.sig
+          gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+          shasum -a 256 -c codecov.SHA256SUM
+          chmod +x codecov
+          ./codecov -F "$(basename $PWD | sed s/[^a-z]/_/g)"
+
+  lint:
+    runs-on: ubuntu-latest
+    container:
+      image: cimg/python:3.10
```
mikewilli: Is it better to run these jobs in a python container vs doing the `actions/setup-python` step like in metric-hub? https://github.com/mozilla/metric-hub/blob/main/.github/workflows/build-docs.yml#L16-L20

mikewilli: Ok, I found the [docs](https://docs.github.com/en/actions/tutorials/build-and-test-code/python#using-the-default-python-version) on this and they recommend using the `setup-python` action instead of the python container so that everything is explicit. Probably doesn't make a huge difference, but we also get caching there so maybe worth changing?

.github/workflows/tagged-deploy.yml
```diff
@@ -0,0 +1,30 @@
+name: Tagged Deploy
+
+on:
+  push:
+    tags:
```
mikewilli: Do we want to include a tags filter or is it ok to run this on every tag?

mikewilli (approved):
Left a few comments, but looks good thanks!

PR #534, opened 2026-04-16
feat: add background-update enrollments query

mikewilli:
Adds a background-update enrollments query that will be used when the config is set to `background-update`.

Follow up to https://github.com/mozilla/metric-hub/pull/1392

Will require a follow-up in Jetstream to use the config value `enrollments_query_type`.


mikewilli:
We should push a new version for this.

scholtzan (approved):

PR #537, opened 2026-06-23
Use _build_exposure_query_glean_events_stream for exposure queries

scholtzan:
Addresses https://github.com/mozilla/jetstream/issues/3186

See discussion in https://mozilla.slack.com/archives/C4D5ZA91B/p1782195139762939, the `events_stream` tables can be used for the exposure queries which results in a lot less data scanned


mikewilli (approved):
LGTM thanks! We should publish a new version as well when this lands.

PR #541, opened 2026-07-02
fix: use actual analysis unit for events stream queries

### LINKED ISSUE #538: Bug: experiment analysis with Glean IDs returns empty results for group-randomized experiments
Running experiment analysis with the Glean-ID option (`--use-glean-ids` in Jetstream) on an experiment that is randomized by **profile group** rather than by individual client silently produces empty / zero values for _every_ metric. The run reports success, so the output is easily mistaken for "no effect" instead of "no data."

The root cause seems to be that the metrics are matched to clients on the **profile_group_id** according to the experiment setup column, but the Glean enrollment (and exposure) query is built on hard-coded **client_id**. The two ID types never match, so every join misses and all metric values collapse to zero/null.

While this seems to only affect profile-group-randomized experiments, this can be more prevalent because that is currently the default that Experimenter assigns for Desktop experiments.

## Reproduction steps

1. Pick a group-randomized experiment, e.g. here `desktop-address-autofill-switzerland-release-holdback`
2. Run the analysis with `--use-glean-ids` (see below). For simplicity, you can run a single date with a single analysis period (one daily window) to demonstrate this cheaply.
3. The run succeeds and the enrollment table is populated, but every metric value comes back
   zero/empty. The same single-window run **without** `--use-glean-ids` returns real values. The
   failure is independent of which metrics are configured (it is upstream of metric selection, in
   the shared enrollment join), so the default metric config is sufficient.

Example command:

```bash
jetstream run \
  --project_id={TESTING_PROJECT_NAME} \
  --dataset_id={TESTING_DATASET_NAME} \
  --bucket=none \
  --experiment_slug=desktop-address-autofill-switzerland-release-holdback \
  --date=2026-01-24 \
  --analysis_periods=day \
  --use-glean-ids
# Single daily window (2026-01-24 is a few days after this experiment's enrollment end).
```

On this test experiment, this showed the day-4 `active_hours` exposures-basis mean was **0.0** with `--use-glean-ids` vs **0.78** without. The other metrics show the same.

## Possible fix

When building the Glean enrollment and exposure queries on the `Normandy` path, pass the
experiment's real analysis unit instead of letting it default to `client_id`. Both builders
already accept the argument, so this is a two-line change in [src/mozanalysis/experiment.py](https://github.com/mozilla/mozanalysis/blob/fd5c8dfe73acd07228d92b6455988023a2cdc236/src/mozanalysis/experiment.py):

```diff
@@ -738,7 +738,8 @@ def _build_enrollments_query(self, ...):
         if enrollments_query_type == EnrollmentsQueryType.NORMANDY:
             if use_glean_ids:
                 return self._build_enrollments_query_glean_events_stream(
-                    time_limits, self.app_id, sample_size
+                    time_limits, self.app_id, sample_size,
+                    analysis_id=self.analysis_unit.value,
                 )
@@ -799,6 +800,7 @@ def _build_exposure_query(self, ...):
             if use_glean_ids:
                 return self._build_exposure_query_glean_events_stream(
                     time_limits,
+                    client_id_field=self.analysis_unit.value,
                 )
```

`self.analysis_unit.value` resolves to `profile_group_id` for group-randomized experiments and to
`client_id` for client-randomized ones, so client-randomized experiments are unchanged.

mikewilli:
fixes #538


mikewilli:
We should push a new tag for version `2026.7.1` when this lands so that it can be updated in jetstream.

scholtzan (approved):

PR #549, opened 2026-07-30
fix: anchor exposures off of enrollment date

mikewilli:


mikewilli:
Decision was made to proceed with anchoring exposures off of the `exposure_date` instead. Closing...

